PR #7, opened 2017-10-17
EZP-28095: Nothing happens when trying to upload a file with greater size than allowed

sunpietro:
https://jira.ez.no/browse/EZP-28095


dew326 (approved):

dew326:
This is not included in the componentized modules. Since we include this bundle in the meta repo, the modules should be componentized and pushed into this bundle.

PR #9, opened 2017-10-18
EZP-28101: Incorrect path to location view in the sub items list item

sunpietro:
https://jira.ez.no/browse/EZP-28101


src/modules/sub-items/sub.items.module.js
```diff
@@ -334,5 +337,6 @@ SubItemsModule.defaultProps = {
         }
     },
     limit: 10,
-    offset: 0
+    offset: 0,
+    locationViewLink: '/admin/content/location/{{locationId}}'
```
dew326: I think this should not be hardcoded here. If we want to hardcode it for now, it should be in the admin-ui with TODO for backend to provide this link

dew326 (changes requested):

dew326 (approved):

PR #12, opened 2017-10-20
EZP-28106: Sub Items List module doesn't list sub items in the Media folder

sunpietro:
https://jira.ez.no/browse/EZP-28106


lserwatka (approved):

dew326 (changes requested):
`yarn run componentize`? 
Without it, this change will not be included in admin-ui

dew326 (approved):

PR #14, opened 2017-11-07
EZP-28196: Improve canSelectContent callback in UDW

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-28196


src/modules/universal-discovery/components/content-meta-preview/content.meta.preview.component.js
```diff
@@ -64,6 +65,21 @@ export default class ContentMetaPreviewComponent extends Component {
     }
 
     /**
+     * Toggles the disable state on select content button
+     *
+     * @method toggleDisableState
+     * @param {Boolean} The disable state
```
sunpietro: I think it should be: `@param {Boolean} disabled The disabled state`

src/modules/universal-discovery/components/content-meta-preview/content.meta.preview.component.js
```diff
@@ -72,13 +88,14 @@ export default class ContentMetaPreviewComponent extends Component {
...
     renderSelectContentBtn() {
         const {data, canSelectContent, onSelectContent, labels} = this.props;
-        const canSelect = canSelectContent(data.ContentInfo.Content);
         const attrs = {
             className: 'c-meta-preview__btn--select',
             onClick: onSelectContent
         };
 
-        if (!canSelect) {
+        canSelectContent(data, this.toggleDisableState.bind(this));
```
sunpietro: I'm still trying to make up an another solution for it. It will run the checking everytime the button is rerendered, but the following code will be executed without waiting fo the results of that check.
I think it might cause issues in the future, because there might be a racing condition.
The button should be rendered after the `canSelectContent` method's callback is invoked, not earlier.

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -301,7 +301,7 @@ UniversalDiscoveryModule.defaultProps = {
     loadContentTypes,
     findContentBySearchQuery,
     findLocationsByParentLocationId,
-    canSelectContent: () => true,
+    canSelectContent: (item, callback) => callback(false),
```
sunpietro: Maybe: `true`? Previously we were always allowing to select a content.

sunpietro (changes requested):
You have to make out another solution. Less error-prone.

src/modules/universal-discovery/universal.discovery.module.js
dew326: It is this case. `Callback` expects boolean if the button should be disabled. `false` means `do not disable button`.

src/modules/universal-discovery/components/content-meta-preview/content.meta.preview.component.js
dew326: The button will be disabled until the `canSelectContent` method will allow enabling the button.

sunpietro (approved):

PR #28, opened 2017-12-15
EZP-28547: SubItems should not redirect to form on edit item

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-28547

**Requires** https://github.com/ezsystems/ezplatform-admin-ui/pull/185


sunpietro (changes requested):
Missing doc in the usage-examples

dew326:
As discussed we should have only one documentation and it's not the one in the repo.

@DominikaK we need to update documentation once this is merged.

sunpietro (approved):

PR #36, opened 2017-12-18
EZP-28588: When Sub Items List module is empty it should not display an empty table state

sunpietro:
…
https://jira.ez.no/browse/EZP-28588


m-tyrala (approved):
QA tests ok.

dew326 (approved):

PR #41, opened 2017-12-20
EZP-28023: As an editor I would like to load more items in UDW when browsing content

sunpietro:
https://jira.ez.no/browse/EZP-28023

load-more-udw


dew326 (approved):

m-tyrala (approved):
QA tests ok. All works fine on Google Chrome. However there're some problems with Firefox (which i consulted with @sunpietro ), that will be fixed in another PR.

PR #43, opened 2017-12-20
EZP-28631: [UDW] Hide pagination when search returns only one page

sunpietro:
https://jira.ez.no/browse/EZP-28631


lserwatka:
@sunpietro could you rebase here?

PR #55, opened 2018-03-02
EZP-28841: Content On the Fly v2

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-28841
**Requires:** https://github.com/ezsystems/ezplatform-admin-ui/pull/366

**TODO**
- [x] content on the fly in UDW
- [x] forced language
- [x] check permission - missing endpoint


src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -0,0 +1,132 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import { loadLocation } from '../../services/universal.discovery.service';
+
+import './css/content.creator.component.css';
+
+export default class ContentCreatorComponent extends Component {
+    constructor(props) {
```
sunpietro, line 9: If you're not using props to initialize state, then you can omit the `props` variable in the `constructor` method.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -0,0 +1,132 @@
...
+
+    handlePublish() {
+        this._iframe.contentWindow.onbeforeunload = () => {};
+        this._iframe.contentWindow.document.body.querySelector('#ezrepoforms_content_edit_publish').click();
+    }
+
+    handleIframeLoad() {
+        const locationId = this._iframe.contentWindow.document.querySelector('meta[name="LocationID"]');
+
+        if (this._iframe.contentWindow.location.pathname !== this.generateIframeUrl() && !locationId) {
```
sunpietro: Can `this.generateIframeUrl()` be cached?

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -0,0 +1,132 @@
...
+
+            return;
+        }
+
+        if (locationId) {
+            this.loadLocationInfo(locationId.content);
+        } else {
+            this.setState(state => Object.assign({}, state, { iframeLoading: false }));
+
+            this._iframe.contentWindow.onbeforeunload = () => {
```
sunpietro: Question: I would put the iframe event handlers to the second callback of `this.setState` method. Do you think it's a bad approach? When moving the event handlers assignment to the callback you apply them right the state is updated (in a sync way).

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -0,0 +1,132 @@
...
+        ));
+
+        promise
+            .then((response) => {
+                this.props.handlePublish(response.View.Result.searchHits.searchHit[0].value.Location);
+            });
+    }
+
+    generateIframeUrl() {
+        return window.Routing.generate('ezplatform.content.create.on_the_fly', {
```
sunpietro: I think it should be initialized in the `constructor` method, because you're not updating app state in case of props updates. This function will always return the same output.
In the `constructor` it should be stored liked this:
```javascript
this.iframeUrl = window.Routing.generate(...);
```

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Renders a loading state icon
+     *
+     * @method renderLoadingIcon
+     * @returns {Element}
+     * @memberof FinderTreeLeafComponent
+     */
+    renderLoadingIcon() {
```
sunpietro: `renderLoadingSpinner`

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -0,0 +1,132 @@
...
+                <div className="m-ud c-content-creator">
+                    <h1 className="m-ud__title">{title}</h1>
+                    <div className="m-ud__content-wrapper">
+                        <div className={contentClass} ref={ref => this._refContentContainer = ref}>
+                            {this.renderLoadingIcon()}
+                            <iframe
+                                src={iframeUrl}
+                                ref={ref => this._iframe = ref}
+                                className="c-content-creator__iframe"
+                                onLoad={this.handleIframeLoad.bind(this)}
```
sunpietro: Please, move all the bindings to the `constructor`, so in the `constructor` you should do like this: `this.handleIframeLoad = this.handleIframeLoad.bind(this)`

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -0,0 +1,132 @@
...
+
+        return (
+            <svg className="ez-icon ez-spin ez-icon-x2 ez-icon-spinner">
+                <use xlinkHref="/bundles/ezplatformadminui/img/ez-icons.svg#spinner"></use>
+            </svg>
+        );
+    }
+
+    render() {
+        const title = this.props.labels.contentOnTheFly.creatingContent
```
sunpietro: Nitpick: You are getting data from the props in many places. Maybe you should do something like this: `{labels, selectedContentType, ...} = this.props;`. It will help you to avoid repetetive writing `this.props` everywhere.

src/modules/universal-discovery/components/content-creator/css/content.creator.component.css
```diff
@@ -0,0 +1,49 @@
+.c-content-creator .m-ud__content-wrapper {
```
sunpietro, line 1: I believe we agreed to move the CSS code directly into SASS files of a bundle. We are not importing CSS into the components anymore.

src/modules/universal-discovery/components/content-creator/css/content.creator.component.css
```diff
@@ -0,0 +1,49 @@
...
+    padding: 0 0 16px 0;
+    grid-template-rows: 1fr 64px;
+    grid-template-areas: "main" "footer";
```
sunpietro: Single quotes, please.

src/modules/universal-discovery/components/content-creator/css/content.creator.component.css
```diff
@@ -0,0 +1,49 @@
...
+}
+
+.c-content-creator__iframe {
+    width: 100%;
+    border: none;
+}
+
+.m-ud__action--publish {
+    font-weight: 700;
+    background: #f15a10;
```
sunpietro, line 29: Recently, I've found out that we're exposing the CSS variables in the `:root` of our styles. Maybe we should start using CSS variables? So `background: var(--orange)`;

src/modules/universal-discovery/components/content-meta-preview/content.meta.preview.component.js
```diff
@@ -93,6 +95,10 @@ export default class ContentMetaPreviewComponent extends Component {
             onClick: onSelectContent
         };
 
+        if (this.props.activeTab === TAB_CREATE) {
```
sunpietro: I would move this conditional to the beginning of the method body.

src/modules/universal-discovery/components/create/create.choose.content.type.component.js
```diff
@@ -0,0 +1,96 @@
...
+        if (this.state.selected.identifier === item.identifier) {
+            attrs.className = `${attrs.className} is-selected`;
+        }
+
+        if (this.state.filterQuery && !item.name.toLowerCase().includes(this.state.filterQuery)) {
+            attrs.hidden = true;
+        }
+
+        return (
+            <div {...attrs} key={index} onClick={this.updateSelectedItem.bind(this, item)}>
```
sunpietro: `onClick={this.updateSelectedItem.bind(this, item)}` can be moved to the `attrs` variable declaration.

src/modules/universal-discovery/components/create/create.choose.content.type.component.js
```diff
@@ -0,0 +1,96 @@
...
+        const value = event.target.value.toLowerCase();
+
+        window.clearTimeout(this._filterTimeout);
+
+        this._filterTimeout = window.setTimeout(() => {
+            this.setState(state => Object.assign({}, state, {filterQuery: value}));
+        }, 200);
+    }
+
+    renderItem(item, index) {
```
sunpietro: To simplify the code here, maybe you should create a stateless item component. The same applies to the stateless group component. Do you agree with this approach?

src/modules/universal-discovery/components/create/create.choose.content.type.component.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    render() {
+        return (
+            <div className="c-choose-content-type">
+                <p className="c-choose-content-type__title">{this.props.labels.contentOnTheFly.selectContentType}</p>
+                <div className="c-choose-content-type__list-wrapper">
+                    <input className="form-control" type="text" placeholder="Type to refine" onChange={this.updateFilterQuery} />
+                    <div className="c-choose-content-type__list" style={{maxHeight:`${this.props.maxHeight - 232}px`}}>
+                        {Object.keys(this.props.contentTypes).map(this.renderGroup)}
```
sunpietro: Instead of `Object.keys` you might want to use `Object.values`

src/modules/universal-discovery/components/create/create.choose.content.type.component.js
```diff
@@ -0,0 +1,96 @@
...
+            </div>
+        );
+    }
+
+    render() {
+        return (
+            <div className="c-choose-content-type">
+                <p className="c-choose-content-type__title">{this.props.labels.contentOnTheFly.selectContentType}</p>
+                <div className="c-choose-content-type__list-wrapper">
+                    <input className="form-control" type="text" placeholder="Type to refine" onChange={this.updateFilterQuery} />
```
sunpietro: Missing label in the `placeholder` attribute?

src/modules/universal-discovery/components/create/create.choose.language.component.js
```diff
@@ -0,0 +1,70 @@
...
+import './css/create.choose.language.component.css';
+
+export default class ChooseLanguageComponent extends Component {
+    constructor(props) {
+        super(props);
+
+        this.updateSelection = this.updateSelection.bind(this);
+
+        this.state = {
+            selected: this.props.languages[0]
```
sunpietro: `props.languages[0]` instead of `this.props.languages[0]`

src/modules/universal-discovery/components/create/create.choose.language.component.js
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    updateSelection(event) {
+        const value = event.target.value;
```
sunpietro: Maybe `languageCode` instead?

src/modules/universal-discovery/components/create/create.choose.language.component.js
```diff
@@ -0,0 +1,70 @@
...
+        this.setState(state => Object.assign({}, state, {selected: selectedLanguage}));
+    }
+
+    renderOption(item, index) {
+        const attrs = {
+            key: index,
+            value: item.languageCode
+        }
+
+        if (item.languageCode === this.props.forcedLanguage) {
```
sunpietro: Question: When do you force using a specific language?

src/modules/universal-discovery/components/create/create.choose.language.component.js
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        return (
+            <option {...attrs}>{item.name}</option>
+        );
+    }
+
+    render() {
+        const selectAttrs = {
+            className: "form-control",
```
sunpietro: Single quotes, please.

src/modules/universal-discovery/components/create/create.choose.language.component.js
```diff
@@ -0,0 +1,70 @@
...
+            onChange: this.updateSelection.bind(this)
```
sunpietro: Binding should be done in the `constructor`

src/modules/universal-discovery/components/create/create.component.js
```diff
@@ -0,0 +1,28 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import ChooseLanguageComponent from './create.choose.language.component';
+import ChooseContentTypeComponent from './create.choose.content.type.component';
+
+import './css/create.component.css';
+
+export default class CreateComponent extends Component {
```
sunpietro: It should be a stateless component

src/modules/universal-discovery/components/finder/finder.component.js
```diff
@@ -236,5 +238,7 @@ FinderComponent.propTypes = {
         token: PropTypes.string.isRequired,
         siteaccess: PropTypes.string.isRequired
     }).isRequired,
-    labels: PropTypes.object.isRequired
+    labels: PropTypes.object.isRequired,
+    allowOnlyContainers: PropTypes.bool,
```
sunpietro: Doesn't `allowContainersOnly` sound better?

src/modules/universal-discovery/components/tab-content/create.panel.component.js
```diff
@@ -0,0 +1,54 @@
...
+
+    if (!props.isVisible) {
+        wrapperAttrs.hidden = true;
+    }
+
+    return (
+        <div {...wrapperAttrs}>
+            <TabContentPanelComponent {...props}>
+                <div className="c-create-panel__first-step">
+                    <div className="c-create-panel__step-title">1) {props.labels.contentOnTheFly.chooseLangaugeAndContentType}</div>
```
sunpietro, line 23: Using `ol > li` isn't the right way to go?

src/modules/universal-discovery/services/universal.discovery.service.js
```diff
@@ -59,6 +59,82 @@ export const loadContentInfo = ({token, siteaccess}, contentId, callback) => {
...
+ * @param {Object} params params hash containing REST config: token, siteaccess, contentTypeIdentifier, languageCode, locationId
+ * @param {Function} callback
+ */
+export const checkCreatePermission = ({
+        token,
+        contentTypeIdentifier,
+        languageCode,
+        locationId
+    }, callback) => {
+    const endpoint = window.Routing.generate('ezplatform.content.create.on_the_fly.has_access', {
```
sunpietro: There's no need to duplicate the names: `window.Routing.generate('ezplatform.content.create.on_the_fly.has_access', { languageCode, contentTypeIdentifier, locationId })`

src/modules/universal-discovery/services/universal.discovery.service.js
```diff
@@ -59,6 +59,82 @@ export const loadContentInfo = ({token, siteaccess}, contentId, callback) => {
...
+        languageCode: languageCode,
+        contentTypeIdentifier: contentTypeIdentifier,
+        locationId: locationId
+    });
+    const request = new Request(endpoint, {
+        method: 'GET',
+        headers: {'X-CSRF-Token': token},
+        mode: 'cors',
```
sunpietro: We decided to switch to proper mode: `same-origin`

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -4,30 +4,49 @@ import PropTypes from 'prop-types';
...
     constructor(props) {
         super(props);
 
+        let selectedLanguage = props.languages[0];
+
+        this.onLanguageSelected = this.onLanguageSelected.bind(this);
+        this.onContentTypeSelected = this.onContentTypeSelected.bind(this);
+        this.handlePublish = this.handlePublish.bind(this);
+
+        if (props.forcedLanguageCOTF) {
```
sunpietro: I believe it should be renamed to something like: `cotfForcedLanguage` or maybe it can be done differently. Because, here youre hardcoding a kind-of-dependency to the COTF which might not exist on a specific instalation.
Do you really need this information here, in UDW module?

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -236,19 +309,85 @@ export default class UniversalDiscoveryModule extends Component {
...
+        checkCreatePermission({
+           token: this.props.restInfo.token,
+           contentTypeIdentifier: this.state.selectedContentType.identifier,
+           languageCode: this.state.selectedLanguage.languageCode,
+           locationId: this.state.contentMeta.id
+        }, (response) => {
+            this.setState(state => Object.assign({}, state, {isCheckingPermission: false}));
+
+            if (this.state.hasPermission !== response.access) {
+               this.setState(state => Object.assign({}, state, {hasPermission: response.access}));
```
sunpietro: I think you should set state once, not twice in the same scope.

sunpietro (changes requested):
There's quite a lof of things to change.

src/modules/universal-discovery/components/content-creator/css/content.creator.component.css
dew326: I thought this was applied only to Page Builder. If we move CSS to admin-ui sass we would always need to make two PR when changing something in modules

src/modules/universal-discovery/components/create/create.choose.content.type.component.js
dew326: I need the keys in the map function

src/modules/universal-discovery/components/create/create.choose.language.component.js
dew326: This is forced in AlloyEditor then creating content on the fly.

src/modules/universal-discovery/universal.discovery.module.js
dew326: COTF is part of a UDW module and the configuration must be set on the UDW and then pass to the COTF. 
`Because, here youre hardcoding a kind-of-dependency to the COTF which might not exist on a specific instalation.` - that's why it has a default value `''` which will prevent from breaking in customers instalations

src/modules/universal-discovery/universal.discovery.module.js
dew326: I'm not sure how this could be in one. We want to change `isCheckingPermission` every time when the request is finished but change the permission state only when have to.

sunpietro: What do you think about it?
```javascript
this.setState(state => Object.assign({}, state, {
    isCheckingPermission: false,
    hasPermission: state.hasPermission !== response.access ? response.access : state.hasPermission
}));
```

dew326: I can try but as far as I know thanks to `immutable` with `Object.assign` this will make rerender of react component and this is what I wanted to avoid in this situation.

sunpietro: You're already doing it, so why doing it twice?

sunpietro: React will compare boolean values (I assume they are boolean). If there's any change then it will cause re-rendering of any related piece of code.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
dew326: You can but I don't think you should.
`Class components should always call the base constructor with props.` from React doc: https://reactjs.org/docs/state-and-lifecycle.html#adding-local-state-to-a-class

sunpietro: This is recommendation, right. We can follow this rule. It's especially useful when you're extending your own component classes. In this case, when you extend base component class it's not that relevant. Because any kind of props won't affect the way the base class works.

dew326: I know it will not cause any problems when extending `React.Component` for now, but if they have this in their doc they may require this at some point in the future.

sunpietro: Ok, let's keep it then. From the currect perspective, it brings no value now, but it's a recommended by React doc.

dew326:
ping @sunpietro

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -0,0 +1,140 @@
...
+        this.handleIframeLoad = this.handleIframeLoad.bind(this);
+        this.handlePublish = this.handlePublish.bind(this);
+
+        this.state = {
+            iframeLoading: true
+        };
+    }
+
+    handlePublish() {
+        this._iframe.contentWindow.onbeforeunload = () => {};
```
sunpietro: With ReactJS we abandoned the rule of adding underscores into protected variables. These will not be exposed in the React code anyway.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
sunpietro: The question is still valid.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -0,0 +1,140 @@
...
+        );
+    }
+
+    render() {
+        const {labels, selectedContentType, selectedLanguage, maxHeight, onCancel} = this.props;
+        const title = labels.contentOnTheFly.creatingContent
+                .replace('{contentType}', selectedContentType.name)
+                .replace('{language}', selectedLanguage.name);
+        const iframeUrl = this.generateIframeUrl();
+        const contentClass = this.state.iframeLoading ? "m-ud__content is-loading" : "m-ud__content";
```
sunpietro: CS: Single quotes, please.

sunpietro: Why not: `m-ud__content--is-loading`?

src/modules/universal-discovery/components/create/create.choose.content.type.component.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    updateSelectedItem(item) {
+        this.props.onContentTypeSelected(item);
+
+        this.setState(state => Object.assign({}, state, {selected: item}));
+    }
+
+    updateFilterQuery(event) {
+        const value = event.target.value.toLowerCase();
```
sunpietro: Have you thought about naming it as `filterQuery`?

src/modules/universal-discovery/components/create/create.choose.content.type.component.js
sunpietro: The question is still valid.

src/modules/universal-discovery/components/create/create.choose.language.component.js
```diff
@@ -0,0 +1,71 @@
...
+export default class ChooseLanguageComponent extends Component {
+    constructor(props) {
+        super(props);
+
+        this.updateSelection = this.updateSelection.bind(this);
+        this.renderOption = this.renderOption.bind(this);
+
+        this.state = {
+            selected: props.languages[0]
+        }
```
sunpietro: CS: Missing `;`

src/modules/universal-discovery/components/create/create.choose.language.component.js
```diff
@@ -0,0 +1,71 @@
...
+
+export default class ChooseLanguageComponent extends Component {
+    constructor(props) {
+        super(props);
+
+        this.updateSelection = this.updateSelection.bind(this);
+        this.renderOption = this.renderOption.bind(this);
+
+        this.state = {
+            selected: props.languages[0]
```
sunpietro: Maybe you should rename it to: `selectedLanguage` instead?

src/modules/universal-discovery/components/create/create.choose.language.component.js
```diff
@@ -0,0 +1,71 @@
...
+        this.props.onLanguageSelected(selectedLanguage);
+
+        this.setState(state => Object.assign({}, state, {selected: selectedLanguage}));
+    }
+
+    renderOption(item, index) {
+        const attrs = {
+            key: index,
+            value: item.languageCode
+        }
```
sunpietro: CS: Missing `;`

src/modules/universal-discovery/components/create/create.component.js
sunpietro: It's still valid.

src/modules/universal-discovery/components/create/css/create.choose.content.type.component.css
```diff
@@ -0,0 +1,38 @@
...
+.c-choose-content-type__group-item {
+    padding-left: 16px;
+    cursor: pointer;
+}
+
+.c-choose-content-type__group-item:hover {
+    background-color: #e1f5ff;
+}
+
+
```
sunpietro: CS: Remove extra empty line here.

src/modules/universal-discovery/components/tab-content/create.panel.component.js
sunpietro: Hmm?

src/modules/universal-discovery/services/universal.discovery.service.js
```diff
@@ -59,6 +59,82 @@ export const loadContentInfo = ({token, siteaccess}, contentId, callback) => {
...
+            'X-CSRF-Token': token
+        }),
+        body,
+        mode: 'same-origin',
+        credentials: 'same-origin'
+    });
+
+    fetch(request)
+        .then(handleRequestResponse)
+        .then(json => callback(json))
```
sunpietro: `.then(callback)`

src/modules/universal-discovery/services/universal.discovery.service.js
```diff
@@ -59,6 +59,82 @@ export const loadContentInfo = ({token, siteaccess}, contentId, callback) => {
...
+    const request = new Request(endpoint, {
+        method: 'GET',
+        headers: {'X-CSRF-Token': token},
+        mode: 'same-origin',
+        credentials: 'same-origin'
+    });
+
+    fetch(request)
+        .then(handleRequestResponse)
+        .then(json => callback(json))
```
sunpietro: `.then(callback)`

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -194,11 +244,32 @@ export default class UniversalDiscoveryModule extends Component {
...
+                onContentTypeSelected: this.onContentTypeSelected,
+                contentTypesMap: this.state.contentTypesMap,
+                forcedLanguage: this.props.cotfForcedLanguage
+            }
+        };
+
+        if (this.props.onlyContentOnTheFly) {
+            return (
+                <div className="m-ud__panels">
+                    {this.renderSinglePanel(createPanelConfig)}
```
sunpietro, line 263: I'm thinking about moving these configs into some kidn of array. But maybe in other PR.

sunpietro (changes requested):

src/modules/universal-discovery/components/create/create.choose.content.type.component.js
dew326: I don't see any point in creating such small components which will we only render HTML.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
dew326: Those `onbeforeunload` and `onunload` is not related to the `setState`, I think callback for `setState` should be used when its depends on the state.

src/modules/universal-discovery/components/create/create.choose.content.type.component.js
sunpietro: There's at least one point. By moving the code to the stateless component:
1. you don't have to think how it's rendered in the parent component,
2. the component manages it's display by itself by handling props,
3. reduced file size of parent component,
4. parent is responsible for rendering groups (of whatever kind, you define imports on your own) but it doesn't have to care how to render it. You declare, you want to render a group and that's it. Group should have stateless children. Children should care about themselves.

I hope, you can see my point now.

dew326: I agree if the component will have some logic for himself but in this case, it will only get the props and return HTML, that's all.

sunpietro: That's how the stateless components work.

src/modules/universal-discovery/components/tab-content/create.panel.component.js
dew326: So `li` would need to have inside a whole `create` or `finder` component? I think it's not the best structure.

src/modules/universal-discovery/components/create/create.choose.content.type.component.js
```diff
@@ -26,12 +26,12 @@ export default class ChooseContentTypeComponent extends Component {
...
 
     updateFilterQuery(event) {
-        const value = event.target.value.toLowerCase();
+        const filterQuery = event.target.value.toLowerCase();
 
         window.clearTimeout(this._filterTimeout);
 
         this._filterTimeout = window.setTimeout(() => {
-            this.setState(state => Object.assign({}, state, {filterQuery: value}));
+            this.setState(state => Object.assign({}, state, {filterQuery: filterQuery}));
```
sunpietro: `{ filterQuery }`

src/modules/universal-discovery/components/create/create.choose.language.component.js
```diff
@@ -21,14 +21,14 @@ export default class ChooseLanguageComponent extends Component {
 
         this.props.onLanguageSelected(selectedLanguage);
 
-        this.setState(state => Object.assign({}, state, {selected: selectedLanguage}));
+        this.setState(state => Object.assign({}, state, {selectedLanguage: selectedLanguage}));
```
sunpietro: `{ selectedLanguage }`

src/modules/universal-discovery/components/create/create.component.js
```diff
@@ -6,15 +6,13 @@ import ChooseContentTypeComponent from './create.choose.content.type.component';
...
-export default class CreateComponent extends Component {
-    render() {
-        return (
-            <div className="c-create" style={{maxHeight:`${this.props.maxHeight}px`}}>
-                <ChooseLanguageComponent {...this.props} />
-                <ChooseContentTypeComponent {...this.props} />
-            </div>
-        );
-    }
+const CreateComponent = (props) => {
```
sunpietro: You can drop import of React component in the first line: ` import React from 'react';`

sunpietro (approved):

PR #60, opened 2018-03-23
EZP-29009: On empty containers SubItems performs 2 fetch requests

sunpietro:
https://jira.ez.no/browse/EZP-29009


dew326 (approved):

andrerom:
+1, missing thing, and I might read this wrong, but like said [in the issue](https://github.com/ezsystems/ezplatform-admin-ui/issues/405) the `loadContentItems` method should stop and return early if there are no search hits, otherwise it loads content item with empty id.

sunpietro:
@andrerom I believe it's not the case. The only situation there's no children in a given location is when you create a new, empty folder. If there's any children inside, then it will have all the required information. Please, correct me if I'm wrong. I can't imagine a situation when it may break.

andrerom:
Before this fix the code was breaking down on this on empty containers, and if there are other cases this might load on empty containers or non container, or thirdparty choose to reuse it, then it will probably happen again.

The code here first loads locations before loading content, however there is no code checking if you got any locations before proceeding.

See issue description, first there is a search for locations, then a borked search for content with empty content id which should have been returning error, but doesn't.

sunpietro:
I don't agree it was breaking down in that case. It was just doing 2 more requests (that were not needed). If backend doesn't provide items on initial state then sub items will present itself as it has no children inside. After adding new ones, the location view page will be reloaded. Finally, after reloading the page the module will have a new initial state with children inside.
So, if a developer doesn't provide any children as the module props - `items` prop, then I won't try to load anything (never).
Still I believe the issue doesn't exist anymore.

andrerom:
Ok, lets go with that then. The main issue is at least fixed.

andrerom (approved):

micszo (approved):
QA Approve

PR #67, opened 2018-05-28
EZP-29251: Expose field types values when selecting content with UDW

sunpietro:
https://jira.ez.no/browse/EZP-29251

**FOR QA:**
The module should behave almost like before. Nothing should be broken.
The only difference are the new content meta preview states: 
- when loading image for preview,
- when image for preview is not available.


src/modules/universal-discovery/components/content-meta-preview/content.meta.preview.component.js
```diff
@@ -10,59 +10,19 @@ export default class ContentMetaPreviewComponent extends Component {
...
-        if (!response.View.Result.count) {
-            return;
-        }
-
-        const version = response.View.Result.searchHits.searchHit[0].value.Content.CurrentVersion.Version;
-        const imageField = version.Fields.field.find(field => field.fieldTypeIdentifier === 'ezimage');
-        const versionLanguages = version.VersionInfo.VersionTranslationInfo.Language;
-        const translations = versionLanguages.map(langauge => this.props.languages.mappings[langauge.languageCode].name);
-        const imageUri = imageField && imageField.fieldValue ? imageField.fieldValue.uri : null;
+    UNSAFE_componentWillReceiveProps(props) {
```
dew326: As far as I know, we have 16.2 and this is not available in this version? 
Besides that I would use `getDerivedStateFromProps`

dew326 (changes requested):

src/modules/universal-discovery/components/content-meta-preview/content.meta.preview.component.js
sunpietro: For some, unknown reason I have React 16.3.2 installed. Everything seems to be working fine. I've also quick checked Alloy Editor and it also works as expected.

lserwatka: We should try upgrade in 2.2 to latest React and Bootstrap.

dew326: We cannot require 16.3 without new tag and tests on assets bundle. 

What we should do:
1. Use old `componentWillReceiveProps`.
2. Upgrade React in assets bundle.
3. Test components and AlloyEditor.

lserwatka: Yes, let's do it as @dew326 suggested.

sunpietro: @lserwatka updating Bootstrap is risky. I remember last time we updated it and it took some time to fix all the issues. If you want to upgrade it, then now is the last moment to do it.

lserwatka: @sunpietro that is why we need extensive QA on top of it. Should do it.

sunpietro: I fully agree with both of you :)

dew326 (approved):

package.json
```diff
@@ -1,32 +1,26 @@
...
-    "babel-preset-react": "^6.24.1",
-    "clean-webpack-plugin": "^0.1.16",
-    "css-loader": "^0.28.5",
-    "eslint-plugin-react": "^7.3.0",
-    "prop-types": "^15.5.10",
-    "style-loader": "^0.18.2",
-    "uglifyjs-webpack-plugin": "^0.4.6",
-    "webpack": "^3.5.5"
-  }
+    "name": "ezplatform-admin-ui-modules",
```
tischsoic, line 2: Was this file formatted by `yarn/npm` or something else (e.g. prettier)? Because if the latter, after running `yarn/npm install` in the future the file will be reformated back to 2 spaces.

sunpietro: It was formatted by Prettier.

tischsoic: I think that original npm's file format should be preserved in case of `package.json` file.

micszo (approved):
Sanities OK.

PR #68, opened 2018-06-18
EZP-29302: UDW displays N/A instead of item's Content Type

tischsoic:
https://jira.ez.no/browse/EZP-29302


src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -169,13 +170,16 @@ export default class UniversalDiscoveryModule extends Component {
         this.props.onConfirm(this.addContentTypeInfo([location]));
     }
 
-    addContentTypeInfo(content) {
-        return content.map((item) => {
-            item = JSON.parse(JSON.stringify(item));
+    addContentTypeInfo(contents) {
```
sunpietro: Why did you change `contents` to an `array`?

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -188,10 +192,11 @@ export default class UniversalDiscoveryModule extends Component {
 
     onItemSelect(contentMeta) {
         const isLocationAllowed = !this.props.cotfAllowedLocations.length || this.props.cotfAllowedLocations.includes(contentMeta.id);
+        const contentMetaWitchContentTypeInfo = this.addContentTypeInfo([contentMeta])[0];
```
sunpietro: Because of your change you're doing a kind of strange thing here. Currently, I don't see any value in changing the way of handling from a single item to an array of items in the `addContentTypeInfo` method

src/modules/universal-discovery/universal.discovery.module.js
sunpietro: `content` is uncountable. Maybe you should use `contentMeta`

src/modules/universal-discovery/universal.discovery.module.js
dew326: `contentMetaWithContentTypeInfo` - you have `witCh`

sunpietro: Let's call the witch hunter ;)

src/modules/universal-discovery/universal.discovery.module.js
dew326: I would use generic `items`

dew326 (approved):

sunpietro (approved):

micszo (approved):
👍

PR #71, opened 2018-06-20
EZP-29319: I want to know that Confirmed items in UDW are clickable

tischsoic:
https://jira.ez.no/browse/EZP-29319


dew326:
I'm not a big fan of unnecessary refactors. It will cause many conflicts when merging a fix up to master.
This ticket is about to add few CSS rules.

lserwatka:
@dew326 is right, let's split this up on two PRs

tischsoic:
@dew326 There were also 2 changes to JS needed. Not only to CSS. 
e.g. I moved `onClick={this.togglePopup.bind(this)}` up in JSX structure in `selected.content.component.js`. Previously you often asked in such cases to refactor `bind` and so on, so I decided that this time I will refactor the whole file so that everything would be OK.
There is also one change in `popup.component.js`, thus I also refactored this component. 
Split this refactoring?

tischsoic:
I think that partial refactoring is good; we should do this when touching some component. It would be rather a bad idea to change e.g. `Object.assign` to `{... ` in every place in the source code at once because there are no tests. So the other option is to refactor on the go.
What do you  think? @lserwatka  @dew326 @sunpietro

sunpietro:
@tischsoic at first, you should focus only on fixing the issue making as little changes as possible. Then in the separate PR, you might do some code refactoring and there we'll discuss your new approach. Is it ok for you?

tischsoic:
@sunpietro I will just remove unnecessary refactoring.

src/modules/universal-discovery/components/selected-content/selected.content.component.js
```diff
@@ -117,14 +117,18 @@ export default class SelectedContentComponent extends Component {
     }
 
     render() {
-        const titles = this.props.items.map((item) => item.ContentInfo.Content.Name).join(', ');
+        const { items } = this.props;
+        const titles = items.map((item) => item.ContentInfo.Content.Name).join(', ');
+        const anyItemSelected = !!items.length;
+        const cssClassOnAnyItemSelected = anyItemSelected ? 'c-selected-content--any-item-selected' : '';
+        const cssClasses = `c-selected-content ${cssClassOnAnyItemSelected}`;
```
tischsoic: I would be better to use something like this: https://www.npmjs.com/package/classnames
What do you think, @sunpietro @dew326 ?

src/modules/common/popup/popup.component.js
```diff
@@ -20,9 +20,12 @@ export default class PopupComponent extends Component {
      * Hides a popup
      *
      * @method hidePopup
+     * @param {SyntheticEvent} event
      * @memberof PopupComponent
      */
-    hidePopup() {
+    hidePopup(event) {
+        event.stopPropagation();
```
sunpietro: What was the reason to stop event propagation?

src/modules/universal-discovery/components/selected-content/selected.content.component.js
```diff
@@ -117,15 +117,21 @@ export default class SelectedContentComponent extends Component {
...
+        const infoCssClasses = `c-selected-content__info ${cssClassOnAnyItemSelected}`;
 
         return (
             <div className="c-selected-content">
                 {this.renderSelectedItems()}
-                <strong className="c-selected-content__title">{this.getTitle()}</strong>
-                {this.renderLimitLabel()}
-                <div className="c-selected-content__content-names" onClick={this.togglePopup.bind(this)}>
-                    {titles.length ? titles : this.props.labels.selectedContent.noConfirmedContent}
+                <div className={infoCssClasses} onClick={this.togglePopup.bind(this)}>
```
sunpietro: Binding should go to constructor

sunpietro (changes requested):

src/modules/common/popup/popup.component.js
tischsoic: Now there is no exact reason for this, but there was during development.

Popup is rendered within `SelectedContentComponent`. 
So there is that chain of events: 
1. you click on popup.
2. `hidePopup` hides popup, and calls `onClose` from `SelectedContentComponent`
3. click event reaches *top div* of `SelectedContentComponent` 

What is important is that in some point of developement there was `togglePopup` as `onClick` on the *top div* of `SelectedContentComponent`. 
So back then event from point 3 was reaching top div, and in turn `togglePopup` was called so popup was reopened again. Because of `togglePopup` being called.

event reaches popup -> hidePopup -> event reaches *top div* -> togglePopup 
hide + toggle === no change

But at the end I moved `onClick` to another div and there is no longer need for `stopPropagation `.

Remove this `stopPropagation`?

sunpietro: Yes, now it's not needed here.

sunpietro (approved):

dew326 (approved):

PR #72, opened 2018-06-21
EZP-29270: Sub-items sorting order not respected by UDW

tischsoic:
https://jira.ez.no/browse/EZP-29270


dew326 (changes requested):
I don't like the approach. You are creating clauses map based on fetched locations (it's 50 locations but it can be changed by integrator). This could be done in a simpler way. Just create one sortClauses just before `findLocationsByParentLocationId()` you have all information there. It's too many calculations on each load.

src/modules/universal-discovery/components/finder/finder.component.js
```diff
@@ -181,15 +182,16 @@ export default class FinderComponent extends Component {
      * Handles loading more items for a selected parent location
      *
      * @method onLoadMore
-     * @param {Number} parentLocationId
+     * @param {Object} parentLocation
      * @memberof FinderComponent
      */
-    onLoadMore(parentLocationId) {
+    onLoadMore(parentLocation) {
```
dew326, line 188: Potential BC break, you can add new parameter but you cannot change the existing one.

dew326 (changes requested):

src/modules/universal-discovery/components/finder/finder.component.js
```diff
@@ -391,7 +417,9 @@ FinderComponent.propTypes = {
     contentTypesMap: PropTypes.object,
     preselectedLocation: PropTypes.number,
     allowedLocations: PropTypes.array,
-    isVisible: PropTypes.bool
+    isVisible: PropTypes.bool,
+    sortFieldMappings: PropTypes.object,
```
sunpietro: Shouldn't it be required?

src/modules/universal-discovery/components/finder/finder.component.js
lserwatka: Is this a public method or API?

sunpietro (commented):
It's not public.

src/modules/universal-discovery/components/finder/finder.component.js
sunpietro: It's not public method. The public API is about existing extension points. This method is not one of them

lserwatka: OK, so we should be fine.

src/modules/universal-discovery/components/finder/finder.component.js
tischsoic: Good point.

sunpietro (approved):

dew326 (approved):

barbaragr (approved):

PR #74, opened 2018-06-22
Return of the PropTypes to grace

tischsoic:
Removed `babel-plugin-transform-react-remove-prop-types` plugin from `.babelrc`.
Fixed some PropTypes which generated errors in JS' console.

**Connected with**:
https://github.com/ezsystems/ezplatform-admin-ui/pull/535


src/modules/sub-items/sub.items.module.js
```diff
@@ -328,7 +328,7 @@ SubItemsModule.propTypes = {
     updateLocationPriority: PropTypes.func,
     activeView: PropTypes.string,
     extraActions: PropTypes.arrayOf(PropTypes.shape({
-        component: PropTypes.element,
+        component: PropTypes.func,
```
sunpietro, line 331: It's not a function it's a React class reference.
Maybe you should implement custom checking? https://reactjs.org/docs/typechecking-with-proptypes.html

src/modules/universal-discovery/components/content-table/content.table.component.js
```diff
@@ -205,7 +205,7 @@ ContentTableComponent.propTypes = {
     perPage: PropTypes.number.isRequired,
     onItemSelect: PropTypes.func.isRequired,
     contentTypesMap: PropTypes.object.isRequired,
-    onItemClick: PropTypes.func.isRequired,
+    onItemClick: PropTypes.func,
```
sunpietro, line 208: Not required anymore?

sunpietro (changes requested):
Almost good :)

src/modules/sub-items/sub.items.module.js
```diff
@@ -315,7 +315,7 @@ export default class SubItemsModule extends Component {
 }
 
 SubItemsModule.propTypes = {
-    parentLocationId: PropTypes.number.isRequired,
+    parentLocationId: PropTypes.string.isRequired,
```
dew326: LocationId is a number as far as I know.

src/modules/multi-file-upload/multi.file.upload.module.js
```diff
@@ -233,7 +233,7 @@ MultiFileUploadModule.propTypes = {
     }).isRequired,
     parentInfo: PropTypes.shape({
         contentTypeIdentifier: PropTypes.string.isRequired,
-        contentTypeId: PropTypes.number.isRequired,
+        contentTypeId: PropTypes.string.isRequired,
```
dew326: contentTypeId is also a number.

src/modules/sub-items/sub.items.module.js
tischsoic: I was thinking about it, I will investigate it further and I will give feedback. :)

src/modules/sub-items/sub.items.module.js
tischsoic: I guest it is number as string. There where errors that `string` is provided instead of `number`. I can check it once more if you want.

sunpietro: Can you enforce passing it as a Number? For instance with `parseInt()`

src/modules/universal-discovery/components/content-table/content.table.component.js
tischsoic: By default it is `null`, thus it cannot be set as `isRequired`.

src/modules/sub-items/sub.items.module.js
dew326: In admin-ui we probably passing string instead of Number.

src/modules/sub-items/sub.items.module.js
tischsoic: I checked it and options depend on what we assume about `component`. If it is:
1. _class component_ - we can check if it inherits from React.Component
2. _functional component_ or _class component_ - we can only check if it is a function (see: https://github.com/facebook/react/issues/9186#issuecomment-289695205)
3. HTML or _2._ - we can check if it is a _function_ or a _string_

tischsoic: @sunpietro What do you think?

src/modules/sub-items/sub.items.module.js
tischsoic: Should I change it in `ezplatform-admin-ui`? Or call `parseInt` somewhere in `universal.discovery.module.js`?

src/modules/sub-items/sub.items.module.js
sunpietro: Let it be of `function` type then

src/modules/sub-items/sub.items.module.js
dew326: Should be done in `ezplatform-admin-ui`.

sunpietro: You should update the place where you're calling sub items module and there parse the value passed to `parentLocationId` as a Number.

sunpietro: Basically, it should be a Number. But fixing it should create a new pull request in admin-ui repo.

mnocon (approved):
Looks good to me!

PR #77, opened 2018-06-27
EZP-29338: Can select content without information about content image in UDW

sunpietro:
https://jira.ez.no/browse/EZP-29338

ezp-29338


micszo (commented):
Test OK but no CR. :)

PR #78, opened 2018-06-27
EZP-29314: Lots of error messages after clicking fast bookmark button

tischsoic:
https://jira.ez.no/browse/EZP-29314


src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -213,6 +214,29 @@ export default class UniversalDiscoveryModule extends Component {
...
+
+    /**
+     * Returns true if bookmark is loading, otherwise returns false
+     *
+     * @method isBookmarkLoading
+     * @param {String} locationId
+     * @memberof UniversalDiscoveryModule
+     */
+    isBookmarkLoading(locationId) {
+        return locationId in this.loadingBookmarksLocationsIds && this.loadingBookmarksLocationsIds[locationId];
```
dew326: I think `!!this.loadingBookmarksLocationsIds[locationId]` is enough.

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -250,22 +286,53 @@ export default class UniversalDiscoveryModule extends Component {
...
     isBookmarked(locationId) {
         const { bookmarked } = this.state;
-        if (locationId in bookmarked) {
-            return bookmarked[locationId];
-        }
+        const locationBookmarkChecked = locationId in bookmarked;
 
-        this.checkIsBookmarked(locationId);
+        if (!locationBookmarkChecked) {
+            this.checkIsBookmarked(locationId);
```
dew326: Now I realize we have method `isBookmarked` and `checkIsBookmarked` little bit confusing.

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -250,22 +286,53 @@ export default class UniversalDiscoveryModule extends Component {
...
+            return null;
```
dew326: And method `isBookmarked` returns `null` it's like: "I don't know" ;)

dew326 (changes requested):

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -52,13 +52,14 @@ export default class UniversalDiscoveryModule extends Component {
...
+        this.toggleBookmark = this.toggleBookmark.bind(this);
         this.onBookmarkAdded = this.onBookmarkAdded.bind(this);
-        this.setBookmarked = this.setBookmarked.bind(this);
+        this.onBookmarkRemoved = this.onBookmarkRemoved.bind(this);
+        this.setBookmarkLoadingStatus = this.setBookmarkLoadingStatus.bind(this);
         this.requireBookmarksCount = this.requireBookmarksCount.bind(this);
         this.onBookmarksLoaded = this.onBookmarksLoaded.bind(this);
 
-        this._refBookmarksPanelComponent = null;
+        this.loadingBookmarksLocationsIds = {};
```
sunpietro: What is the idea behind this property?

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -221,25 +245,37 @@ export default class UniversalDiscoveryModule extends Component {
      * @memberof UniversalDiscoveryModule
      */
     checkIsBookmarked(locationId) {
+        if (this.isBookmarkLoading(locationId)) {
```
sunpietro: I think you should update state, then re-render a bookmark with loading state.

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -445,17 +512,16 @@ export default class UniversalDiscoveryModule extends Component {
         }
 
         const { contentTypesMap, maxHeight, activeTab, contentMeta } = this.state;
-        const isSelectedContentBookmarked = this.isBookmarked(contentMeta.id);
+        const isContentMetaBookmarked = this.isBookmarked(contentMeta.id);
```
sunpietro: I don't like the new name. It states whether the content meta is bookmarked, not the content.

sunpietro (changes requested):

src/modules/universal-discovery/universal.discovery.module.js
tischsoic: Because it returns `null` to say "I don't know". :P
When the response from server has not yet arrived, you do not know if it is bookmarked or not. ;)

src/modules/universal-discovery/universal.discovery.module.js
tischsoic: Good point

src/modules/universal-discovery/universal.discovery.module.js
sunpietro: IMO, there should be a promise, that updates state of bookmarked flag at the end. If we don't know, then it should return `false`.

sunpietro: Otherwise, we'll end up like video detection in the browsers. `Maybe we support this video format, but who knows?`

src/modules/universal-discovery/universal.discovery.module.js
tischsoic: What do you suggest instead?

src/modules/universal-discovery/universal.discovery.module.js
dew326: I like the promise to wait for the answer than returning "I don't know" or false.

src/modules/universal-discovery/universal.discovery.module.js
tischsoic: Currently when bookmark is loading there is no bookmark being displayed to user.
The problem is that render function should be pure and you should not update state during rendering, otherwise React shows warnings. This is the reason, why I decided to store ids of loading bookmarks in: `this.loadingBookmarksLocationsIds = {};`.

src/modules/universal-discovery/universal.discovery.module.js
dew326: `checkIsBookmarked` to`fetchBookmarkedStatus`?

src/modules/universal-discovery/universal.discovery.module.js
tischsoic: I don't like this approach. Elements should be rendered based on values from _state_ and _props_. This is how the cycle should look like in React. 
I don't know where you would like to put this Promise and how it would suppose to work. 
Maybe I don't understand what you are suggesting. :)

src/modules/universal-discovery/universal.discovery.module.js
tischsoic: See response to your comment below.

src/modules/universal-discovery/universal.discovery.module.js
sunpietro: I mean, default state is `false`. If content is bookmarked, then it will be changed to `true` when promise resolves.

lserwatka:
@tischsoic rebase is needed here

lserwatka:
@dew326 @sunpietro are you ok with this change?

dew326 (approved):

sunpietro (approved):

barbaragr (approved):

PR #79, opened 2018-06-29
EZP-29360: Bookmark button in UDW is placed too low

tischsoic:
https://jira.ez.no/browse/EZP-29360


src/modules/universal-discovery/components/content-meta-preview/css/content.meta.preview.component.css
```diff
@@ -130,6 +130,7 @@
 }
 
 .c-meta-preview__content-bookmark .ez-add-to-bookmarks__icon-wrapper {
+    padding: 0rem;
```
dew326: `0` should be enough

dew326 (changes requested):

dew326 (approved):

barbaragr (approved):

PR #80, opened 2018-06-29
EZP-29155: Priority buttons work in the same way

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29155
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no


dew326 (changes requested):
This is not a valid solution.
We have a bug but somewhere else. `onChange` we take the value from input value (which is static as React doesn't update those values). 
So for the fix, you have to change in this line:
https://github.com/ezsystems/ezplatform-admin-ui-modules/blob/master/src/modules/sub-items/components/table-view/table.view.item.component.js#L81
`this._refPriorityInput.value` into `event.target.value`

lserwatka:
@ViniTou do you have any progress on this one?

ViniTou:
@lserwatka 
We were discussing this internally on slack, solution provided by @dew326 doesn't work, and as was decided with @sunpietro this solution is acceptable in this case.

dew326 (approved):

lserwatka:
Great, let's keep the ball rolling

barbaragr (approved):

PR #81, opened 2018-06-29
EZP-29362: Error when deleting bookmark from UDW

tischsoic:
https://jira.ez.no/browse/EZP-29362

Fixed error on removing bookmark when in bookmarks tab in UDW.


barbaragr (approved):

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -325,7 +325,7 @@ export default class UniversalDiscoveryModule extends Component {
         const { activeTab } = this.state;
 
         if (activeTab === TAB_BOOKMARKS) {
-            this.onItemSelect(null);
+            this.closeCMP();
```
sunpietro, line 328: It should be named like `this.closeContentMetaPreview`. Please, use meaningful method/function names.

PR #82, opened 2018-07-02
UDW code improvements

sunpietro:
UDW code improvements


src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -218,10 +214,7 @@ export default class UniversalDiscoveryModule extends Component {
      * @param {String} id
      */
     onItemRemove(id) {
-        this.setState((state) => ({
-            ...state,
-            selectedContent: state.selectedContent.filter((item) => item.id !== id),
-        }));
+        this.setState((state) => ({ ...state, selectedContent: state.selectedContent.filter((item) => item.id !== id) }));
```
tischsoic, line 221: For me one, such long line is unreadable.

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -361,18 +353,14 @@ export default class UniversalDiscoveryModule extends Component {
...
-        items.forEach((bookmarkLocation) => {
-            this.setBookmarked(bookmarkLocation.Location.id, true);
-        });
+        this.setState(
+            (state) => ({
+                ...state,
+                bookmarksDuringLoadingCount: state.bookmarksDuringLoadingCount - items.length,
+                userBookmarksItems: [...items, ...state.userBookmarksItems],
+            }),
+            () => items.forEach((bookmarkLocation) => this.setBookmarked(bookmarkLocation.Location.id, true))
```
tischsoic, line 366: For me, when it is split in 2 lines it is way more readable.

tischsoic (approved):

src/modules/universal-discovery/universal.discovery.module.js
sunpietro: Using arrow functions is a way to improve readability of single liners. If you and @dew326 agree that it's too long, then I'm fine to split it into multiple lines :)

dew326: Don't have we prettier configured with: `"printWidth": 140`? If so I would stick to this, because there might be some problems with auto correct in the future.

dew326 (approved):

src/modules/universal-discovery/universal.discovery.module.js
tischsoic: @dew326 The probability that Prettier will change it's formatting practices is weak. It currently allows a developer to decide how this line should be formatted and nothing suggests that it is going to change. Suggesting that, because we set 140 limit, we should try to make lines of code that long is completely wrong and harmful to the code.

dew326: What I meant was the line length should be automatically corrected by prettier, not manual by us. If prettier decided this line is less than 140 it's ok for me.

sunpietro: I'd stick with current formating, because thanks to it we won't discuss CS issues anymore. 
In development you can set whatever spacing you want, but the final PR, please format it to the standards that the team agreed.

tischsoic: OK, I agree, I'm really sorry @dew326.
 I checked how Prettier formats code in this case. I thought there is option to split it in reasonable way to two lines. Sometimes prettier allows it, but not in this case.

tischsoic: What I meant is that Prettier also allows you to format this code in this way:
```
() => {
    items.forEach((bookmarkLocation) => {
        this.setBookmarked(bookmarkLocation.Location.id, true);
    });
}
```
So it's not true that Prettier output will always be the same for the same code.

barbaragr (approved):

PR #83, opened 2018-07-03
Implemented more performant deep clone helper

sunpietro:
This is another pack of code improvements. The behaviour of Universal Discovery should not differ from the behaviour on master.


src/modules/common/helpers/deep.clone.helper.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * Clones any object. Faster alternative to `JSON.parse(JSON.stringify)`
+ *
+ * @function deepClone
+ * @param {Any} data
+ * @returns {Any} cloned data
+ */
+const deepClone = (data) => {
```
tischsoic, line 8: Did you make some speed tests? I have made few simple tests and it seems that the speeds are basicaly the same.
But these were quite simple speed tests. :)

tischsoic: My simple test suite: https://jsbin.com/zetanapoha/edit?js,output 🏎 :)

sunpietro: This is the same approach we're using in Page Builder. I wanted to make it inline.

tischsoic: > inline

I don't know what you mean by that?

tischsoic: It is just to make it behave the same way as in Page Builder?

sunpietro: Exactly

tischsoic: I think that deep clone output should always be the same, why not use previous approach from UDW in Page Builder? It is much simpler and it seems that performance is the same. But nevermind. :)

sunpietro: I did other performance tests and it told me that the previous approach was faster only in FF. In other browsers this approach was better.
We can discuss it in private and, eventually, prepare test cases for each option :)

sunpietro:
This branch has been rebased with master

src/modules/common/helpers/deep.clone.helper.js
tischsoic: If you performed more detailed tests and your approach performed better then OK. :)
Maybe in the future will find some time to test it more thoroughly. :P

tischsoic (approved):

barbaragr (approved):

PR #84, opened 2018-07-03
EZP-29381: Selected item becomes unselected for a moment while browsing content in UDW

sunpietro:
https://jira.ez.no/browse/EZP-29381


src/modules/universal-discovery/components/finder/finder.component.js
```diff
@@ -3,63 +3,80 @@ import PropTypes from 'prop-types';
...
         }
     }
 
     componentDidUpdate() {
         this.updateBranchesContainerScroll();
-    }
 
-    componentWillReceiveProps(nextProps) {
-        const isShowingUp = nextProps.isVisible && nextProps.isVisible !== this.props.isVisible;
+        const isShowingUp = this.props.isVisible;
```
dew326: This variable should reflect if the component was not visible and now is visible. I think `this.props.isVisible` is not that case.

src/modules/universal-discovery/components/finder/finder.tree.branch.component.js
```diff
@@ -10,17 +10,20 @@ export default class FinderTreeBranchComponent extends Component {
...
+            currentlyLoadingLocationId: null,
+            selectedLocationId: null,
         };
     }
 
-    componentWillReceiveProps(props) {
-        this.setState(state => Object.assign({}, state, {
-            selectedLocations: props.selectedLocations,
-            currentlyLoadingLocationId: false
+    UNSAFE_componentWillReceiveProps(props) {
```
tischsoic: I think it would be good idea to start using new lifecycle methods, in this case `getDerivedStateFromProps`.

src/modules/universal-discovery/components/selected-content/selected.content.component.js
```diff
@@ -10,15 +10,14 @@ export default class SelectedContentComponent extends Component {
...
-
+        this.hidePopup = this.hidePopup.bind(this);
         this.togglePopup = this.togglePopup.bind(this);
+        this.renderSelectedItem = this.renderSelectedItem.bind(this);
+
+        this.state = { items: [], isPopupVisible: false };
     }
 
-    componentWillReceiveProps(props) {
+    UNSAFE_componentWillReceiveProps(props) {
```
tischsoic: I think it would be good idea to start using new lifecycle methods, in this case `getDerivedStateFromProps`.

src/modules/universal-discovery/components/finder/finder.tree.leaf.component.js
```diff
@@ -52,36 +32,41 @@ export default class FinderTreeLeafComponent extends Component {
...
-        const hasChildren = location.childCount ? `${componentClassName}--has-children` : '';
-        const isLoadingChildren = this.state.isLoadingChildren ? `${componentClassName}--loading` : '';
-        const finalClassName = `${componentClassName} ${isSelected} ${hasChildren} ${isLoadingChildren} ${isNotSelectable}`;
+        const isSelectedClassName = selected ? `${componentClassName}--selected` : '';
+        const isNotSelectableClassName = !isSelectable || isForcedLocation ? `${componentClassName}--not-selectable` : '';
+        const hasChildrenClassName = location.childCount ? `${componentClassName}--has-children` : '';
+        const isLoadingChildrenClassName = isLoadingChildren ? `${componentClassName}--loading` : '';
         const attrs = {
-            className: finalClassName,
+            className: [
```
tischsoic: It would be cleaner to use package like _classnames_; as far as I remember, there are other places where we can use it. Unless we do not want to add 500B to the size of the module.
Reference for the size data: https://bundlephobia.com/result?p=classnames@2.2.6

tischsoic: E.g. `render` method from `finder.tree.branch.component.js` uses `if`s to create className.

src/modules/universal-discovery/components/selected-content/selected.content.component.js
sunpietro: Right, but in this case I just wanted to turn off the linting errors. Feel free to convert this method into a static one.

dew326 (changes requested):

sunpietro:
ping @dew326

src/modules/universal-discovery/components/finder/finder.tree.branch.component.js
```diff
@@ -32,21 +38,31 @@ export default class FinderTreeBranchComponent extends Component {
...
-            });
-        });
-
-        this.props.onItemClick({
-            parent: location.id,
-            location
-        });
+        this.setState(
+            (state) => {
+                return {
```
tischsoic: I think we do not need `return`:
```js
(state) => ({
   ...state,
   //...
})
```
?

dew326 (approved):

tischsoic (approved):

barbaragr (approved):

PR #86, opened 2018-07-04
Coding standards fixes

sunpietro:
Coding standards fixes. Fixed using Prettier. That avoid commiting CS fixes in the future pull requests and it will help us focusing on real fixes.

- [x] fixed CS using Prettier,
- [x] Moved from `Object.assign` to object spread in Sub Items List,
- [x] Corrected binding context in the Sub Items List

PR #89, opened 2018-07-05
Run Prettify on every js file

tischsoic:
Everything should work as before.
Prettiefy was run in every file.


dew326:
Duplicated? https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/86

PR #92, opened 2018-07-13
Code improvements

sunpietro:
## Things covered
- [x] bindings corrected,
- [x] removed copying component full state while updating a state,
- [x] Converted into stateful components:
  - popup component,
  - grid view component,
  - load more component,
  - sub items list component,
  - view swticher component
- [x] removed `UNSAFE_componentWillReceiveProps` lifecycle method from some components,
- [x] corrected error handling.

## For QA
The following modules should be retested:
- [x] Universal Discovery,
- [x] Sub Items List,
- [x] Multi File Upload.


src/modules/sub-items/components/load-more/load.more.component.js
```diff
@@ -36,25 +32,25 @@ export default class LoadMoreComponent extends Component {
...
         const delta = totalCount - loadedCount;
         const loadMoreLimit = delta > limit ? limit : delta;
-        const actionText = this.props.labels.action.replace('{{limit}}', loadMoreLimit);
+        const actionText = labels.action.replace('{{limit}}', loadMoreLimit);
 
         return (
             <div className="c-load-more">
                 <div className="c-load-more__message" dangerouslySetInnerHTML={{ __html: loadMoreText }} />
-                <div {...btnAttrs} ref={(ref) => (this._refLoadMore = ref)}>
+                <div {...btnAttrs} ref={this.setLoadMoreRef}>
```
tischsoic, line 66: Do we need `_refLoadMore`? I can see that we use this reference only in one place - in `loadMore` - and only to check if `disabled` flag was set, but this flag is set in the component so in the component we have information if button is disabled thus we don't need reference.

I think that it would be a good idea to extract this: `!totalCount || loadedCount >= totalCount` into another function and use this new function to check this condition in `render` method and in `loadMore` method.

src/modules/sub-items/components/sub-items-list/sub.items.list.component.js
```diff
@@ -1,34 +1,21 @@
...
-            },
-        };
-    }
-
-    UNSAFE_componentWillReceiveProps(props) {
-        this.setState((state) => ({ ...state, activeView: props.activeView }));
-    }
+const views = {
+    table: TableViewComponent,
+    grid: GridViewComponent,
```
tischsoic, line 11: Maybe it would be a good idea to extract `table` and `grid` keys to external const variables, because we use them here and in other place - we get one of these keys from `props`.

e.g.:
```js
export const VIEW_KEY_TABLE = 'table';
```

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -41,7 +41,7 @@ export default class TableViewComponent extends Component {
         const sortKey = this.state.sortKey;
 
         if (!sortKey) {
-            this.setState((state) => ({ ...state, items }));
+            this.setState(() => ({ items }));
```
tischsoic, line 44: We do not need callback here.
```js
this.setState({ items });
```

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -51,7 +51,7 @@ export default class TableViewComponent extends Component {
         } else if (sortKey === SORTKEY_PRIORITY) {
             this.sortByPriority(items);
         } else {
-            this.setState((state) => ({ ...state, items }));
+            this.setState(() => ({ items }));
```
tischsoic, line 54: We do not need callback here.

src/modules/sub-items/components/table-view/table.view.item.component.js
```diff
@@ -31,7 +31,7 @@ export default class TableViewItemComponent extends Component {
      * @memberof TableViewItemComponent
      */
     enablePriorityInput() {
-        this.setState((state) => ({ ...state, priorityInputEnabled: true }));
+        this.setState(() => ({ priorityInputEnabled: true }));
```
tischsoic, line 34: We do not need callback here and in a few other places in this file and in other modified files.

tischsoic (changes requested):

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -78,7 +78,6 @@ export default class TableViewComponent extends Component {
             });
 
             return {
-                ...state,
```
tischsoic, line 81: I know that it might be outside of the scope of this task but methods: `sortByName`, `sortByDate`, `sortByPriority` are basically identical. What changes is only sorting function. We can extract overlaping parts to another function.

I was also thinking about rewriting sort function from this:
```js
items.sort((a, b) => {
    if (isAscSort) {
        return a.content.Name.localeCompare(b.content.Name);
    }
    return b.content.Name.localeCompare(a.content.Name);
});
```
into this:
```js
items.sort((a, b) => {
    [a, b] = isAscSort ? [a, b] : [b, a];
   
    return a.content.Name.localeCompare(b.content.Name);
});
```
to reduce code duplication. I think that it is also more clear.

Sorry for this random thoughts. :)

tischsoic: We may also extract swapping logic to some helper function to make it even more clear:
```js
items.sort((a, b) => {
    [a, b] = swap([a, b], isAscSort);
   
    return a.content.Name.localeCompare(b.content.Name);
});
```

tischsoic: I was thinking about it yesterday and I would gladly improve this piece of code on my own in another PR, if you don't mind. I would like to try using memoization to improve performance. :)

src/modules/sub-items/components/sub-items-list/sub.items.list.component.js
sunpietro: I don't see any point in doing it right now. Can you give me an example in our current codebase, when it will be useful?

src/modules/sub-items/components/table-view/table.view.component.js
sunpietro: Nope, we're going to use callbacks here because we want to make the state updates asynchronous.

src/modules/sub-items/components/table-view/table.view.component.js
sunpietro: See comment above: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/92#discussion_r202540313

src/modules/sub-items/components/table-view/table.view.item.component.js
sunpietro: See comment above: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/92#discussion_r202540313

src/modules/sub-items/components/table-view/table.view.component.js
sunpietro: Good idea. But we don't make to make any big changes. If you feel that you will need such helpers in the future, then you're free to make such improvements. Right now, this is out of scope of this PR.

src/modules/sub-items/components/load-more/load.more.component.js
sunpietro: This idea is out of scope of this PR.

src/modules/sub-items/components/table-view/table.view.component.js
tischsoic: As far as I know, they are always asynchronous.

src/modules/sub-items/components/sub-items-list/sub.items.list.component.js
tischsoic: In `/src/modules/sub-items/components/view-switcher/view.switcher.component.js` in line 76 we have: 
```js
{this.renderButton('table', 'view-list', listViewBtnLabel)}
{this.renderButton('grid', 'view-grid', gridViewBtnLabel)}
```
It is good practice to extract common constants to one shared variable.

tischsoic: We also have in `/src/modules/sub-items/sub.items.module.js` in line 372: `activeView: 'table',`

src/modules/sub-items/components/table-view/table.view.component.js
sunpietro: I'm not so sure

src/modules/sub-items/components/sub-items-list/sub.items.list.component.js
sunpietro: In our current codebase it's not needed. I know, you're trying to follow someone's good practices but we have our own standards. In redux, there's a pattern to expose the action names as constants. That's fine. But we don't need it here, as it's not going to change in the lifecycle of this app. In the future, if there's a need to change it, then we'll change it. I'm not a big fan of premature optimizations.

barbaragr (approved):

dew326 (approved):

lserwatka:
@dew326 @tischsoic can we merge this one safely?

lserwatka:
@sunpietro could you rebase and include missing translations. These were already merged.

sunpietro:
@lserwatka branch has been rebased. @barbaragr unfortunately, it has to be retested once again, please.

barbaragr (approved):

barbaragr:
Looks fine :)

tischsoic (approved):

PR #94, opened 2018-07-18
EZP-29182/EZP-29187: Improvements in Universal Discovery Widget single and multi selection

dew326:
**Jira tickets:** https://jira.ez.no/browse/EZP-29187, https://jira.ez.no/browse/EZP-29182

**Screenshots**
multiple-browse
multiple-search


src/modules/universal-discovery/components/content-table/content.table.item.component.js
```diff
@@ -1,33 +1,48 @@
...
+
 import './css/content.table.item.component.css';
 
-const ContentTableItemComponent = (props) => {
-    const { onItemClick, onPreview, data, contentTypesMap, labels } = props;
-    const item = data.ContentInfo.Content;
-    const contentType = contentTypesMap ? contentTypesMap[item.ContentType._href] : false;
-    const contentTypeName = contentType ? contentType.names.value[0]['#text'] : labels.contentTableItem.notAvailable;
-    const onClick = !!onItemClick ? onItemClick.bind(null, data) : null;
+export default class ContentTableItemComponent extends Component {
```
sunpietro: You should go back to stateless component. The `renderSelectContentBtn` can be defined outside of `const ContentTableItemComponent` definition, just pass the props to it.

src/modules/universal-discovery/components/finder/finder.tree.branch.component.js
```diff
@@ -92,9 +102,7 @@ export default class FinderTreeBranchComponent extends Component {
         }
 
         return (
-            <button
-                className="c-finder-tree-branch__load-more"
-                onClick={() => this.props.onLoadMore(this.props.parentLocation)}>
+            <button className="c-finder-tree-branch__load-more" onClick={() => this.props.onLoadMore(this.props.parentLocation)}>
```
sunpietro: Please, create a class method for handling `onClick` it wil prevent additional re-renderings on each update.

src/modules/universal-discovery/components/select-content-button/css/select.content.button.component.css
```diff
@@ -0,0 +1,14 @@
+.c-select-content-button {
+	position: absolute;
+	right: 1.3rem;
+	top: 50%;
```
sunpietro: It shouldn't be here. Because it's an information about block position. It should go to parent container style definitions.

src/modules/universal-discovery/components/select-content-button/select.content.button.component.js
```diff
@@ -0,0 +1,95 @@
...
+    toggleEnabledState(selectContentEnabled) {
+        if (this.state.selectContentEnabled === selectContentEnabled) {
+            return;
+        }
+
+        this.setState((state) => ({ ...state, selectContentEnabled }));
+    }
+
+    handleSelect(event) {
+        event.stopPropagation();
```
sunpietro, line 42: Why not `preventDefault()`?

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -175,6 +187,13 @@ export default class UniversalDiscoveryModule extends Component {
         this.props.onConfirm(this.addContentTypeInfo(this.state.selectedContent));
     }
 
+    handleSingleConfirm() {
+        this.setState(
+            (state) => ({ ...state, selectedContent: [this.state.contentMeta] }),
```
sunpietro: `(state) => ({ selectedContent: [state.contentMeta] }),`

sunpietro (changes requested):

lserwatka:
@dew326 did you cover bookmarks as well?

src/modules/universal-discovery/components/content-table/content.table.item.component.js
dew326: Good catch. During development I changed this to component with state and then removed the state :)

dew326:
@lserwatka Yes, I covered everything. `Browse`, `Search`, `Create`, `Bookmarks` in both `single` and `multiple` selection. Just didn't want to attach too many screenshots.

lserwatka:
@dew326 ok, clear. Well done!

src/modules/universal-discovery/components/select-content-button/select.content.button.component.js
dew326: Because I don't want to bubble this event to the parent.

tischsoic (approved):

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -106,6 +108,16 @@ export default class UniversalDiscoveryModule extends Component {
         if (!!contentMeta && !isPreviewMetaReady) {
             this.props.loadContentInfo(this.props.restInfo, contentMeta.ContentInfo.Content._id, this.updateContentMetaWithCurrentVersion);
         }
+
+        if (!this.props.multiple) {
+            this.canSelectContent(contentMeta, (canSelectContent) => {
```
sunpietro: Do you think it would be a good idea to extract the callback into a method like `setCanSelectContentState`?

sunpietro (approved):

barbaragr:
1. According to the specification UDW title should be:
- ‘Select content item(s)’ for multi selection
- ‘Select a content item’ for single selection 
Now there is: `Select content`

2. Error message `Cannot read property 'identifier' of undefined`
- configure `vendor/ezsystems/ezplatform-admin-ui/src/bundle/Resources/config/universal_discovery_widget.yml`: for default configuration in `allowed_content_types` type article
`allowed_content_types: [article]`
- clear cache
- create Cotent Type with ezojcetrelationlist
- go to Content/Content structure and click Create and choose CT from previous step
- click Select content - UDW is opened for a while, then it disappear and error message is shown: `Cannot read property 'identifier' of undefined`

3. Scenario `2.2.2.8 Content item’s name link`  from spec https://ezno.sharepoint.com/:w:/r/sites/ProductTeam/_layouts/15/Doc.aspx?sourcedoc=%7B137050e3-dd3f-4629-913f-e9e1f2c1dfce%7D&action=default doesn't work at all. Is it included in this PR's scope? Similar question for next scenario `Content item’s parent name link` - there is no parent name in search results...

barbaragr (changes requested):
4. While embedding a content item in a rich text field, when I launch the UDW to select a content item, title should be: `Select a content **item** to embed` and now there is: `Select a content to embed` (according to scenario 2.2.1.2 from https://ezno.sharepoint.com/:w:/r/sites/ProductTeam/_layouts/15/Doc.aspx?sourcedoc=%7B0623259a-d7fa-4365-9807-207e44697563%7D&action=default)

5. Scenario 2.2.1.3 Title of the UDW when adding an image should be `Select an image` and now there is `Select an image to embed` 

6. Scenario 2.2.2.3 Disabling the ‘Confirm’ button   - there is no tooltip for disabled Confirm button. Tooltips message required by scenario: `The Content Type is not allowed for selection`

7. Group of scenarios 2.2.3 - similar like in point 3 - is it included in this PR?

dew326:
All points with the incorrect title are out of the scope of this PR (1, 4, 5).
3, 7 - switching to browse while searching is also out of scope. This can be done with "large content structure"

Only 2 and 6 are valid.

dew326:
@barbaragr fixed

barbaragr (approved):

PR #107, opened 2018-10-17
EZP-28067; EZP-29713: Implement bulk delete, bulk move

tischsoic:
## JIRA
- https://jira.ez.no/browse/EZP-28067 - bulk delete
- https://jira.ez.no/browse/EZP-29713 - bulk move

## Depends on
- https://github.com/ezsystems/ezplatform-admin-ui/pull/691
- https://github.com/ezsystems/ezplatform-admin-ui/pull/702 

_(**for QA**: on branch https://github.com/ezsystems/ezplatform-admin-ui/pull/702 there are also changes from https://github.com/ezsystems/ezplatform-admin-ui/pull/691)_

## Screenshots

screen shot 2018-10-17 at 12 35 40

screen shot 2018-10-30 at 09 47 13

screen shot 2018-10-30 at 09 48 24

screen shot 2018-10-17 at 12 35 50

screen shot 2018-10-17 at 12 36 09


src/modules/common/popup/popup.component.js
```diff
@@ -1,38 +1,199 @@
...
+            this.props.onClose();
+        }
+    }
+
+    setModalRef(component) {
+        this._refModal = component;
+    }
+
+    renderHeader() {
+        const closeBtnLabel = Translator.trans(/*@Desc("Close")*/ 'popup.close.label', {}, 'universal_discovery_widget');
```
tischsoic, line 83: ping @sunpietro @dew326 Do we want to have separate translations domain for `common` components?

src/modules/common/popup/css/popup.component.css
```diff
@@ -1,39 +1,90 @@
-.c-popup {
-    background: #ededed;
-    padding: 16px;
-    box-shadow: 0 0 30px 0 rgba(0,0,0,.35);
+body.ezs-non-scrollable {
```
sunpietro: WE should remove this from the file. Because we already have a generic CSS class defined in `_general.scss` - `.ez-scroll-disabled`. We should use this one instead. Can you correct it, please?

src/modules/common/popup/css/popup.component.css
```diff
@@ -1,39 +1,90 @@
...
+    overflow: hidden;
     position: relative;
-    border-radius: 4px;
+}
+
+/* .c-popup__dialog {
```
sunpietro: Commented code has to be removed

src/modules/common/popup/css/popup.component.css
```diff
@@ -1,39 +1,90 @@
...
+    min-height: 2.5rem;
+    background: #ededed;
+}
+
+.c-popup__footer--plain {
+    border-top: none;
+}
+
+.c-popup__body {
+    /* height: calc(100% - 60px); */
```
sunpietro: Commented code.

src/modules/common/popup/popup.component.js
sunpietro: I'm not sure we have any.

src/modules/common/popup/popup.component.js
```diff
@@ -1,38 +1,199 @@
...
+
+        const { plain } = this.props;
```
sunpietro: What does `plain` mean in the context of popup component?

src/modules/common/popup/popup.component.js
```diff
@@ -1,38 +1,199 @@
...
+        const { footerChildren } = this.props;
+
+        if (!footerChildren) {
+            return;
+        }
+
+        const { plain } = this.props;
+        let className = 'modal-footer c-popup__footer';
+
+        className = plain ? `${className} c-popup__footer--plain` : className;
```
sunpietro: If it's `plain` then it should be styled by parent container `.c-popup--plain .c-popup__footer`

Resources/translations/sub_items.en.xliff
```diff
@@ -6,6 +6,41 @@
...
+        <target state="new">Cancel</target>
+        <note>key: bulk_delete.popup.cancel</note>
+      </trans-unit>
+      <trans-unit id="a668a561527cc0e3e2c7952fbf300558d9d4f877" resname="bulk_delete.popup.confirm">
+        <source>Send to trash</source>
+        <target state="new">Send to trash</target>
+        <note>key: bulk_delete.popup.confirm</note>
+      </trans-unit>
+      <trans-unit id="02d93b03435046c6bcf08ce15ae7970f7ff8f746" resname="bulk_delete.popup.message">
+        <source>Are you sure you want to send the selected content item(s) to the trash?</source>
```
DominikaK: `<source>Are you sure you want to send the selected content item(s) to trash?</source>`
`to trash`, for consistency with other strings.

src/modules/common/tooltip-popup/css/tooltip.popup.component.css
```diff
@@ -0,0 +1,39 @@
+.c-tooltip-popup {
+    background: #ededed;
```
sunpietro, line 2: Bootstrap exposes CSS variables in the scope of `document`. Maybe we should create our own CSS variables in the Admin UI and reuse them in the other projects. What's your take on it? @dew326

Resources/translations/sub_items.en.xliff
tischsoic: Good pick, @DominikaK, indeed even in inVision prototypes there is `to trash`. My mistake. :-)

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,163 @@
...
+        this.closePopup = this.closePopup.bind(this);
+        this.onPopupConfirm = this.onPopupConfirm.bind(this);
+
+        this.state = {
+            isPopupVisible: false,
+        };
+    }
+
+    componentDidMount() {
+        this.modalContainer = document.createElement('div');
```
sunpietro: I think it's a good practice to define this variable in the `constructor` and fill it with a value here, like you did it.
It will inform us that there are class-scoped additional variables to be reused in other places.

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,163 @@
...
+
+    onBtnClick() {
+        this.togglePopup(true);
+    }
+
+    bulkDelete() {
+        const { restInfo, selectedItems: itemsToDelete, removeItemsFromList } = this.props;
+        const contentsToDelete = itemsToDelete.map((item) => item.content);
+
+        bulkDeleteContents(restInfo, contentsToDelete, (deletedContents, notDeleted) => {
```
sunpietro: The callback should be exposed as a class method, for better code readability.

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,163 @@
...
+        const cancelLabel = Translator.trans(/*@Desc("Cancel")*/ 'bulk_delete.popup.cancel', {}, 'sub_items');
+        const confirmLabel = Translator.trans(/*@Desc("Send to trash")*/ 'bulk_delete.popup.confirm', {}, 'sub_items');
+
+        return (
+            <Fragment>
+                <button onClick={this.closePopup} type="button" className="btn btn-secondary btn--no" data-dismiss="modal">
+                    {cancelLabel}
+                </button>
+                <button
+                    onClick={this.onPopupConfirm}
```
sunpietro:
```suggestion
                    onClick={this.onPopupConfirm}
                    type="button"
```

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,163 @@
...
+        );
+
+        return ReactDOM.createPortal(
+            <Popup
+                onClose={this.closePopup}
+                isVisible={isPopupVisible}
+                isLoading={false}
+                size="medium"
+                footerChildren={this.renderConfirmationPopupFooter()}
+                plain
```
sunpietro: Is it correct?

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,163 @@
...
+        const disabled = !selectedItems.length;
+        const baseClassName = 'm-sub-items__btn';
+        let className = `${baseClassName} ${baseClassName}--trash`;
+
+        className = disabled ? `${className} ${baseClassName}--disabled` : className;
+
+        return (
+            <Fragment>
+                <div className={className} title={label} onClick={this.onBtnClick}>
+                    <svg className="ez-icon">
```
sunpietro: You should use `ez-icon--medium` and remove CSS properties related to the icon dimensions, as well.

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
```diff
@@ -0,0 +1,135 @@
...
+
+    onBtnClick() {
+        this.toggleUdw(true);
+    }
+
+    bulkMove(location) {
+        const { restInfo, selectedItems: itemsToMove, removeItemsFromList } = this.props;
+        const contentsToMove = itemsToMove.map((item) => item.location);
+
+        bulkMoveLocations(restInfo, contentsToMove, location, (movedLocations, notMoved) => {
```
sunpietro: Please extract the callback.

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
```diff
@@ -0,0 +1,135 @@
...
+        const disabled = !selectedItems.length;
+        const baseClassName = 'm-sub-items__btn';
+        let className = `${baseClassName} ${baseClassName}--move`;
+
+        className = disabled ? `${className} ${baseClassName}--disabled` : className;
+
+        return (
+            <Fragment>
+                <div className={className} title={label} onClick={this.onBtnClick}>
+                    <svg className="ez-icon">
```
sunpietro: You need to apply `ez-icon--medium` CSS class here, do you?

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -150,11 +150,11 @@ export default class TableViewComponent extends Component {
      *
      * @method renderItem
      * @param {Object} data
-     * @returns {Element}
+     * @returns {JSX.Element}
```
sunpietro, line 166: I haven't met such notation before. Do you have any resource regarding it?

src/modules/sub-items/css/sub.items.module.css
```diff
@@ -44,9 +48,37 @@
...
+    font-weight: bold;
+    transition: background 0.3s ease-in-out;
+    align-items: center;
+    height: 40px;
+    width: 40px;
+    justify-content: center;
+    margin-right: 8px;
+}
+
+.m-sub-items__btn .ez-icon {
```
sunpietro: Please, use existing icon's CSS classes :)

src/modules/sub-items/sub.items.module.js
```diff
@@ -256,12 +266,44 @@ export default class SubItemsModule extends Component {
         this.setState(() => ({ activeView }));
     }
 
+    onItemSelect(contentId, isSelected) {
+        const { items } = this.state;
+        const itemIdx = items.findIndex((item) => item.content._id == contentId);
```
sunpietro: What does `itemIdx` mean? Please, use meaningful variable names. The code should be readable like a text in book.

sunpietro (changes requested):

src/modules/common/tooltip-popup/css/tooltip.popup.component.css
tischsoic: Maybe we will start using SCSS in `admin-ui-modules`?

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
tischsoic: Good point!

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
tischsoic: Yes, `attributeName` is shorthand for `attributeName={true}`.

src/modules/common/popup/popup.component.js
tischsoic: I decided to name popup, which does not have header and its footer doesn't have top border a "plain popup". You prefer other name/solution?

src/modules/common/popup/popup.component.js
tischsoic: We have this one! 😄

tischsoic: It's probably the first and the only one, for now. :-)

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
dew326: Honestly, I don't like it. It's completely different from the object shorthand `{ plain }`.
IMO it's less readable.

src/modules/common/popup/popup.component.js
dew326: I think better would be just `noHeader` or something like this. 
And you are adding class twice one to the header and footer. I think better is add class to the parent `c-popup--no-header` and style header and footer.

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,163 @@
...
+    bulkDelete() {
+        const { restInfo, selectedItems: itemsToDelete, removeItemsFromList } = this.props;
+        const contentsToDelete = itemsToDelete.map((item) => item.content);
+
+        bulkDeleteContents(restInfo, contentsToDelete, (deletedContents, notDeleted) => {
+            const deletedContentsIds = deletedContents.map((content) => content._id);
+
+            removeItemsFromList((item) => deletedContentsIds.includes(item.content._id));
+
+            if (deletedContents.length > 0) {
```
dew326: `deletedContents.length` is enough

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,163 @@
...
+                const message = Translator.trans(
+                    /*@Desc("The selected content item(s) have been sent to trash")*/ 'bulk_delete.success.message',
+                    {},
+                    'sub_items'
+                );
+
+                window.eZ.helpers.notification.showSuccessNotification(message);
+            }
+
+            if (notDeleted.length > 0) {
```
dew326: `notDeleted.length`

src/modules/sub-items/services/bulk.service.js
```diff
@@ -0,0 +1,75 @@
...
+        },
+        body: body,
+        mode: 'cors',
+        credentials: 'same-origin',
+    });
+
+    fetch(request)
+        .then(handleRequestResponse)
+        .then(callback)
+        .catch(() => window.eZ.helpers.notification.showErrorNotification('Bulk request failed'));
```
dew326: translation

src/modules/sub-items/sub.items.module.js
```diff
@@ -256,12 +266,44 @@ export default class SubItemsModule extends Component {
...
+    getSelectedItems() {
+        const { items } = this.state;
+
+        return items.filter((item) => item.isSelected);
+    }
+
+    removeItemsFromList(predicate) {
+        this.setState((state) => {
+            const { items } = state;
+            const nextItems = items.filter((item) => !predicate(item));
```
dew326: I don't know what it does. `nextItems` and `predicate` are not meaningful names.

dew326 (changes requested):

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,163 @@
...
+            removeItemsFromList((item) => deletedContentsIds.includes(item.content._id));
+
+            if (deletedContents.length > 0) {
+                const message = Translator.trans(
+                    /*@Desc("The selected content item(s) have been sent to trash")*/ 'bulk_delete.success.message',
+                    {},
+                    'sub_items'
+                );
+
+                window.eZ.helpers.notification.showSuccessNotification(message);
```
dew326: As far as I understand the service there might be a situation when both `deletedContents` and `notDeleted` have items. Displaying two notifications:
- `The selected content item(s) have been sent to trash`
- `You do not have permission to delete at least 1 of the selected content item(s). Please contact your Administrator to obtain permissions.`

Might be confusing for the users.

tischsoic: Yes, I was planning to write about this to PMs.

src/modules/common/popup/popup.component.js
sunpietro: I have the same opinion as @dew326

src/modules/sub-items/components/table-view/table.view.component.js
tischsoic: With `Element` I have underlined in red the whole JSX, because JSX syntax does not generate  `Element` but rather `ReactNode`s. `JSX.Element` is supported by VS Code I'm not sure about other editors.

tischsoic:
ping @sunpietro @dew326

tischsoic:
ping @dew326 @sunpietro

src/modules/common/popup/css/popup.component.css
```diff
@@ -1,39 +1,60 @@
...
+    z-index: 2;
+    width: 56px;
+    height: 56px;
 }
 
 .c-popup__header {
-    position: relative;
+    display: block;
+    background: #dbdbdb;
+    font-family: 'Noto Sans', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, sans-serif;
```
dew326: Why we have to change font-family for this popup header? Shouldn't we have globally fonts?

src/modules/common/popup/popup.component.js
```diff
@@ -1,38 +1,191 @@
...
+        };
+
+        document.body.classList.toggle(CLASS_MODAL_OPEN, isVisible);
+        document.body.classList.toggle(CLASS_NON_SCROLLABLE, isVisible);
+
+        if (additionalClasses) {
+            modalAttrs.className = `${modalAttrs.className} ${additionalClasses}`;
+        }
+
+        modalAttrs.className = noHeader ? `${modalAttrs.className} c-popup--no-header` : modalAttrs.className;
```
dew326: Please use one style `if (condition) {}` or `ternary operator`. For consistency I would use `if () {}`.

src/modules/common/tooltip-popup/css/tooltip.popup.component.css
dew326: I don't like the idea of using CSS variables from bootstrap. We have SCSS and we should aim to support it here as well, but this is scope for another story.

src/modules/common/tooltip-popup/css/tooltip.popup.component.css
```diff
@@ -0,0 +1,39 @@
...
+    padding: 16px;
+    box-shadow: 0 0 30px 0 rgba(0,0,0,.35);
```
dew326: `rgba(0,0,0,.35)` I think this is not our standard. Can you use Prettier on this file?

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,170 @@
...
+        bulkDeleteContents(restInfo, locationsToDelete, this.afterBulkDelete);
+    }
+
+    afterBulkDelete(deletedLocations, notDeleted) {
+        const { removeItemsFromList } = this.props;
+        const deletedLocationsIds = new Set(deletedLocations.map((location) => location.id));
+
+        removeItemsFromList(deletedLocationsIds);
+
+        if (deletedLocations.length) {
```
dew326: In some cases, we will display both notifications: `The selected content item(s) have been sent to trash` and `You do not have permission to delete at least 1 of the selected content item(s). Please contact your Administrator to obtain permissions.`. This is confusing for the user.

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,170 @@
...
+                /*@Desc("You do not have permission to delete at least 1 of the selected content item(s). Please contact your Administrator to obtain permissions.")*/ 'bulk_delete.error.message',
+                {},
+                'sub_items'
+            );
+
+            window.eZ.helpers.notification.showErrorNotification(message);
+        }
+    }
+
+    togglePopup(show) {
```
dew326: Nitpick: you can change `show` to `isPopupVisible` and use just `{ isPopupVisible }`

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,170 @@
...
+        const label = Translator.trans(/*@Desc("Delete selected items")*/ 'trash_btn.label', {}, 'sub_items');
+        const disabled = !selectedItems.length;
+        const baseClassName = 'm-sub-items__btn';
+        let className = `${baseClassName} ${baseClassName}--trash`;
+
+        className = disabled ? `${className} ${baseClassName}--disabled` : className;
+
+        return (
+            <Fragment>
+                <div className={className} title={label} onClick={this.onBtnClick}>
```
dew326: Why `div` not `button`?

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,170 @@
...
+                    <svg className="ez-icon ez-icon--medium">
```
dew326: You can use the `Icon` component

src/modules/sub-items/components/bulk-delete/css/bulk.delete.btn.css
```diff
@@ -0,0 +1,13 @@
+.m-sub-items__btn--trash {
```
dew326: Not the correct place for this CSS?

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,170 @@
...
+        return ReactDOM.createPortal(
+            <Popup
+                onClose={this.closePopup}
+                isVisible={isPopupVisible}
+                isLoading={false}
+                size="medium"
+                footerChildren={this.renderConfirmationPopupFooter()}
+                noHeader={true}
+            >
+                <div className="c-bulk-delete-btn__modal-body">{confirmationMessage}</div>
```
dew326: I think everything in `Popup` should have class `c-bulk-delete-popup` not `btn`.

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
```diff
@@ -0,0 +1,138 @@
...
+        bulkMoveLocations(restInfo, contentsToMove, location._href, this.afterBulkMove.bind(this, location));
+    }
+
+    afterBulkMove(location, movedLocations, notMoved) {
+        const { removeItemsFromList } = this.props;
+        const movedLocationsIds = new Set(movedLocations.map((location) => location.id));
+
+        removeItemsFromList(movedLocationsIds);
+
+        if (notMoved.length) {
```
dew326: Again there will be both notifications `success` and `fail`.

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
```diff
@@ -0,0 +1,138 @@
...
+                /*@Desc("The selected content item(s) have been sent to <u>%location_name%</u>")*/ 'bulk_move.success.message',
+                { location_name: location.ContentInfo.Content.Name },
+                'sub_items'
+            );
+
+            window.eZ.helpers.notification.showSuccessNotification(message);
+        }
+    }
+
+    toggleUdw(show) {
```
dew326: nitpick: you can change this to `isUdwOpened`

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
```diff
@@ -0,0 +1,138 @@
...
+        const udwProps = {
+            title,
+            restInfo,
+            onCancel: this.closeUdw,
+            onConfirm: this.onUdwConfirm,
+            canSelectContent: ({ item }, callback) => {
+                callback(!excludedMoveLocations.includes(item.id));
+            },
+            multiple: false,
+            allowContainersOnly: true,
```
dew326: I wonder if this UDW should also have some configuration from the backend. I think it should.

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
```diff
@@ -0,0 +1,138 @@
...
+        const disabled = !selectedItems.length;
+        const baseClassName = 'm-sub-items__btn';
+        let className = `${baseClassName} ${baseClassName}--move`;
+
+        className = disabled ? `${className} ${baseClassName}--disabled` : className;
+
+        return (
+            <Fragment>
+                <div className={className} title={label} onClick={this.onBtnClick}>
+                    <svg className="ez-icon ez-icon--medium">
```
dew326: `Icon` component?

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
```diff
@@ -0,0 +1,138 @@
...
+        const label = Translator.trans(/*@Desc("Move selected items")*/ 'move_btn.label', {}, 'sub_items');
+        const disabled = !selectedItems.length;
+        const baseClassName = 'm-sub-items__btn';
+        let className = `${baseClassName} ${baseClassName}--move`;
+
+        className = disabled ? `${className} ${baseClassName}--disabled` : className;
+
+        return (
+            <Fragment>
+                <div className={className} title={label} onClick={this.onBtnClick}>
```
dew326: Why not `button`.

src/modules/sub-items/components/bulk-move/css/bulk.move.btn.css
```diff
@@ -0,0 +1,4 @@
+.m-sub-items__btn--move {
```
dew326: Probably not the correct place for this CSS.

src/modules/sub-items/components/table-view/table.view.item.component.js
```diff
@@ -163,8 +161,19 @@ export default class TableViewItemComponent extends Component {
         );
     }
 
+    /**
+     *
```
dew326, line 166: Missing doc

src/modules/sub-items/services/bulk.service.js
```diff
@@ -0,0 +1,77 @@
+import { handleRequestResponse } from '../../common/helpers/request.helper.js';
+
+const HEADERS_BULK = {
+    Accept: 'application/vnd.ez.api.BulkOperationResponse+json',
+    'Content-Type': 'application/vnd.ez.api.BulkOperation+json',
+};
+
+export const bulkMoveLocations = (restInfo, locations, newLocationHref, callback) => {
+    const requestBody = getBulkMoveRequestOperations(locations, newLocationHref);
```
dew326: It's not `requestBody`. `operations`?

src/modules/sub-items/services/bulk.service.js
```diff
@@ -0,0 +1,77 @@
...
+
+    makeBulkRequest(restInfo, requestBody, processBulkResponse.bind(null, locations, 201, callback));
+};
+
+export const bulkDeleteContents = (restInfo, locations, callback) => {
+    bulkMoveLocations(restInfo, locations, '/api/ezp/v2/content/trash', callback);
```
dew326: It would be nice to put `'/api/ezp/v2/content/trash'` in some `CONST` on top of this file.

src/modules/sub-items/services/bulk.service.js
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    return operations;
+};
+
+const processBulkResponse = (locations, successCode, callback, response) => {
+    const { operations } = response.BulkOperationResponse;
+    const contentsMatches = { success: [], fail: [] };
+
+    for (const locationId of Object.keys(operations)) {
```
dew326: I think it would be better to use 
```js
Object.entries(operations).forEach(([locationId, response])) {...}
```

src/modules/sub-items/services/bulk.service.js
```diff
@@ -0,0 +1,77 @@
...
+        const response = operations[locationId];
+        const respectiveItem = locations.find((location) => String(location.id) === locationId);
```
dew326: The location id is an integer. You should `parseInt` on a string not to convert the integer to string.

src/modules/sub-items/services/bulk.service.js
```diff
@@ -0,0 +1,77 @@
...
+            },
+        };
+    });
+
+    return operations;
+};
+
+const processBulkResponse = (locations, successCode, callback, response) => {
+    const { operations } = response.BulkOperationResponse;
+    const contentsMatches = { success: [], fail: [] };
```
dew326: All operations are on locations, so this should be `locationsMatches`.

src/modules/sub-items/services/bulk.service.js
```diff
@@ -0,0 +1,77 @@
...
+    'Content-Type': 'application/vnd.ez.api.BulkOperation+json',
+};
+
+export const bulkMoveLocations = (restInfo, locations, newLocationHref, callback) => {
+    const requestBody = getBulkMoveRequestOperations(locations, newLocationHref);
+
+    makeBulkRequest(restInfo, requestBody, processBulkResponse.bind(null, locations, 201, callback));
+};
+
+export const bulkDeleteContents = (restInfo, locations, callback) => {
```
dew326: `bulkDeleteLocations`

src/modules/sub-items/services/bulk.service.js
```diff
@@ -0,0 +1,77 @@
...
+        headers: {
+            ...HEADERS_BULK,
+            'X-CSRF-Token': token,
+        },
+        body: JSON.stringify({
+            bulkOperations: {
+                operations,
+            },
+        }),
+        mode: 'cors',
```
dew326: We use `'same-origin'`.

src/modules/sub-items/sub.items.module.js
```diff
@@ -256,12 +267,78 @@ export default class SubItemsModule extends Component {
...
+    getSelectedItems() {
+        const { items, selectedLocationsIds } = this.state;
+        const selectedLocationsIdsArray = [...selectedLocationsIds];
+        const selectedItems = items.filter((item) => selectedLocationsIdsArray.includes(item.location.id));
+
+        return selectedItems;
+    }
+
+    /**
+     * @param {Set} locationsToRemoveIds
```
dew326, line 346: doc should be full. Missing doc here and there.

src/modules/sub-items/sub.items.module.js
```diff
@@ -256,12 +267,78 @@ export default class SubItemsModule extends Component {
...
+     */
+    removeItemsFromList(locationsToRemoveIds) {
+        this.setState((state) => {
+            const { items } = state;
+            const filteredItems = items.filter((item) => !locationsToRemoveIds.has(item.location.id));
```
dew326: `filteredItems` is not a meaningful name.

webpack.common.js
```diff
@@ -46,7 +46,13 @@ module.exports = {
             commonjs2: 'prop-types',
             commonjs: 'prop-types',
             amd: 'prop-types',
-        }
+        },
+        jquery: {
+            root: '$',
```
dew326: It's not `$` it's `jQuery`.

dew326 (changes requested):

webpack.common.js
tischsoic: I've taken it from: https://github.com/ezsystems/ezplatform-page-builder/blob/master/src/bundle/ui-dev/webpack.common.js#L43

dew326: I know, but it turns out that it works by accident.
I've found out about that when I had to debug why form builder is not working in content on the fly.
https://github.com/ezsystems/ezplatform-form-builder/pull/104/files#diff-b3fe84cb611246d5b214be1f657ea94bR49

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
tischsoic: This is scope for another task: https://jira.ez.no/browse/EZP-29747

webpack.common.js
tischsoic: OK, then I will fix this. Do we have some task to fix this in PB?

src/modules/sub-items/services/bulk.service.js
tischsoic: I've taken it from: https://github.com/ezsystems/ezplatform-admin-ui-modules/blob/master/src/modules/sub-items/services/sub.items.service.js
It might be good when switching to something to change it everywhere, or at least write it down somewhere, otherwise it's hard to keep track of changes in our internal standards.

src/modules/common/popup/css/popup.component.css
tischsoic: This file was compied from: https://github.com/ezsystems/ezplatform-page-builder/blob/master/src/bundle/Resources/public/scss/ui/page-builder/components/config-popup.scss
There is SCSS variable: `font-family: $font-family-sans-serif;`, here I replaced it with variable value.

dew326: I'm wondering why we need to override this. Inheritance doesn't work?

webpack.common.js
dew326: I don't think so.

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
tischsoic: For consistency with Multi file upload button: https://github.com/ezsystems/ezplatform-admin-ui-modules/blob/a38b88a96ed6813b0220d9bed7f2055691e6b97e/src/modules/multi-file-upload/multi.file.upload.module.js#L187

src/modules/sub-items/components/bulk-delete/css/bulk.delete.btn.css
tischsoic: I placed it here because it is strictly connected with button from BulkDelete, I will move this to Sub-items CSSes.

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
tischsoic: Which properties should be placed on the backend side?

tischsoic: I'm looking at: https://github.com/ezsystems/ezplatform-admin-ui/blob/9ea409a5c7f8442a752a94e83d05628f5469144f/src/bundle/Resources/public/js/scripts/admin.location.view.js#L73
Should we just pass here some properties for config, or should it be taken from some data-attribute?

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
tischsoic: The same reason as in the BulkDelete case.

src/modules/sub-items/services/bulk.service.js
tischsoic: 👍 I thought we're not using it yet. Will change this. :-)

dew326: Modern browsers support it.

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
dew326: The container of sub-items should have some data-attribute like `data-udw-config` with configuration `single_container` (the same one which is used on `move button` in right sidebar) and this configuration from data-attribute should be passed as a prop.

src/modules/sub-items/services/bulk.service.js
tischsoic: It doesn't matter here, and I thought that using `String(...)` is simpler. I can change it to `parseInt` if you prefer it. 😄

dew326: Probably will work the same but it has more sense to change into the type which variable actual is not to something else.

src/modules/common/popup/css/popup.component.css
tischsoic: @dew326, hah, I appreciate your in-depth review, but I just moved this file from Page Builder. :D
I will check it. 😝

dew326: I know, this should be done when the code was originally introduced but nothing is perfect :P

src/modules/sub-items/services/bulk.service.js
tischsoic: I will change it to `requestBodyOperations`.

dew326: Maybe just `requestOperations` as method is called `getBulkMoveRequestOperations` ;)

src/modules/common/popup/css/popup.component.css
tischsoic: I've checked it, it seems that we can delete it. Good pick! :-)

dew326: Nice, thank you.

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
tischsoic: On second thought I prefer `show` because `togglePopup(isPopupVisible)` suggest that you should pass current state of popup visibility.

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
tischsoic: So you want me to create new component? Because AFAIK we follow rule that class `c-component-name` is strictly connected with component name.

webpack.common.js
tischsoic: https://jira.ez.no/browse/EZEE-2542

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
tischsoic: The same as in https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/107/files#r228485670

src/modules/sub-items/sub.items.module.js
tischsoic: I was thinking for a while for better name and I decided to use `itemsWithoutItemsToRemove`, but it's a bit long, do you have something better in your mind?

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
tischsoic: Implemented in: https://github.com/ezsystems/ezplatform-admin-ui/pull/702/commits/f656f6773ebfb427ec7d854d3fed5559b22142ec
In PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/702

tischsoic:
ping @dew326

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
```diff
@@ -0,0 +1,138 @@
...
+            return null;
+        }
+
+        const UniversalDiscovery = window.eZ.modules.UniversalDiscovery;
+        const { restInfo, parentLocationId, selectedItems, udwConfig } = this.props;
+        const selectedItemsLocationsIds = selectedItems.map((item) => item.location.id);
+        const excludedMoveLocations = [parentLocationId, ...selectedItemsLocationsIds];
+        const title = Translator.trans(/*@Desc("Choose location")*/ 'udw.choose_location.title', {}, 'sub_items');
+        const udwProps = {
+            ...udwConfig,
```
dew326: The `udwConfig` should override the configuration from JS. This is the convention we use. The backend configuration has the priority.

src/modules/sub-items/sub.items.module.js
```diff
@@ -358,6 +472,7 @@ SubItemsModule.propTypes = {
     contentTypesMap: PropTypes.object,
     totalCount: PropTypes.number,
     languages: PropTypes.object,
+    udwConfig: PropTypes.object.isRequired,
```
dew326: As suggested in another bundle, this `udwConfig` name should be more precise.

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
dew326: Well, if so why couple lines below there is: `const baseClassName = 'm-sub-items__btn';`?

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
dew326: `multiple: false,` and `allowContainersOnly: true,` will not be needed when use `single_container` config.

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
```diff
@@ -0,0 +1,138 @@
...
+        const baseClassName = 'm-sub-items__btn';
+        let className = `${baseClassName} ${baseClassName}--move`;
+
+        className = disabled ? `${className} ${baseClassName}--disabled` : className;
+
+        return (
+            <Fragment>
+                <div className={className} title={label} onClick={this.onBtnClick}>
+                    <svg className="ez-icon ez-icon--medium">
+                        <Icon name="move" extraClasses="ez-icon--medium" />
```
dew326: `Icon` inside `svg`? this will be `svg` inside `svg`.

src/modules/sub-items/services/bulk.service.js
```diff
@@ -0,0 +1,77 @@
+import { handleRequestResponse } from '../../common/helpers/request.helper.js';
+
+const HEADERS_BULK = {
+    Accept: 'application/vnd.ez.api.BulkOperationResponse+json',
+    'Content-Type': 'application/vnd.ez.api.BulkOperation+json',
+};
+const TRASH_FAKE_LOCATION = '/api/ezp/v2/content/trash';
```
dew326, line 7: Nice! 👍

dew326 (changes requested):

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
tischsoic: You're right! I was too fast. :)

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
tischsoic: In Sub-items header there are 3 buttons which share the same CSS class, so I reused it here in order not to duplicate code, but I also see your point. 
Do you want different solution? For me current is not perfect but is OK as I didn't come up with anything better.

dew326: I don't have strong opinion on this. Maybe it's ok to stay with this.
I just have mixed feelings when I read this class name: Button has a modal body 🤔

src/modules/sub-items/components/bulk-move/bulk.move.btn.js
```diff
@@ -0,0 +1,137 @@
...
+            return null;
+        }
+
+        const UniversalDiscovery = window.eZ.modules.UniversalDiscovery;
+        const { restInfo, parentLocationId, selectedItems, udwConfigMove } = this.props;
+        const selectedItemsLocationsIds = selectedItems.map((item) => item.location.id);
+        const excludedMoveLocations = [parentLocationId, ...selectedItemsLocationsIds];
+        const title = Translator.trans(/*@Desc("Choose location")*/ 'udw.choose_location.title', {}, 'sub_items');
+        const udwProps = {
+            ...udwConfigMove,
```
dew326: This should override our configuration. The backend configuration has priority.

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
tischsoic: I agree, we may wait for @sunpietro with this. :)

dew326 (approved):

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
```diff
@@ -0,0 +1,169 @@
...
+
+    onBtnClick() {
+        this.togglePopup(true);
+    }
+
+    bulkDelete() {
+        const { restInfo, selectedItems: itemsToDelete } = this.props;
+        const locationsToDelete = itemsToDelete.map((item) => item.location);
+
+        bulkMoveLocationsToTrash(restInfo, locationsToDelete, this.afterBulkDelete);
```
sunpietro: I'm not sure whether the child component should be responsible for that action. I mean, the child component should not touch services. The architecture I proposed requires running any service actions by the parent container component. So there's a single way of providing new state from the top to bottom. The `bulkDelete` action should run a callback from a parent SubItems Module to run the service function.

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
sunpietro: Your comments made me think. I believe the component should be redesigned. I mean, the base component that adds a button to the header of sub items list should be a kind of container that composes a real button that is rendered in the header and a popup. The button itself should not contain a popup, becuase it's a strange architecture.

sunpietro (changes requested):

lserwatka:
Guys, please keep amount of changes as little as possible, we are fare behind with this story. We should start QA as from tomorrow latest to deliver this story.

tischsoic:
regarding: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/107#discussion_r228834882

Don't we have similar architecture here: https://github.com/ezsystems/ezplatform-admin-ui-modules/blob/17de174da0d0df0928faaa6cf217c4eef9064f40/src/modules/multi-file-upload/multi.file.upload.module.js#L221 ? ping @sunpietro

sunpietro:
@tischsoic MFU is a separate, independent component.

src/modules/sub-items/components/bulk-delete/bulk.delete.btn.js
tischsoic: So you want me to move this logic to `ezplatform-admin-ui-modules/src/modules/sub-items/sub.items.module.js`?

sunpietro: I believe that is the correct approach.

tischsoic:
ping @sunpietro

src/modules/common/popup/popup.component.js
```diff
@@ -171,9 +171,9 @@ Popup.propTypes = {
     isVisible: PropTypes.bool,
     isLoading: PropTypes.bool,
     onClose: PropTypes.func.isRequired,
-    onConfigIframeLoad: PropTypes.func.isRequired,
+    onConfigIframeLoad: PropTypes.func,
```
sunpietro, line 174: I would also create dummy function as a `defaultProp` value

src/modules/sub-items/sub.items.module.js
```diff
@@ -335,6 +360,188 @@ export default class SubItemsModule extends Component {
...
+
+    bulkMove(location) {
+        const { restInfo } = this.props;
+        const itemsToMove = this.getSelectedItems();
+        const contentsToMove = itemsToMove.map((item) => item.location);
+
+        bulkMoveLocations(restInfo, contentsToMove, location._href, this.afterBulkMove.bind(this, location));
+    }
+
+    afterBulkMove(location, movedLocations, notMoved) {
```
sunpietro: I would rename `notMoved` to `itemsNotMoved`

src/modules/sub-items/sub.items.module.js
```diff
@@ -335,6 +360,188 @@ export default class SubItemsModule extends Component {
...
+        this.removeItemsFromList(movedLocationsIds);
+
+        if (notMoved.length) {
+            const message = Translator.trans(
+                /*@Desc("You do not have permission to move at least 1 of the selected content item(s). Please contact your Administrator to obtain permissions.")*/ 'bulk_move.error.message',
+                {},
+                'sub_items'
+            );
+
+            window.eZ.helpers.notification.showErrorNotification(message);
```
sunpietro, line 398: It should display a message and proceed further, right? Or should it break with `return`?

sunpietro (changes requested):
Just a couple of minor issues left.

src/modules/sub-items/sub.items.module.js
tischsoic: It should proceed further.

tischsoic: Because it should display both messages for now; it is going to be changed in https://jira.ez.no/browse/EZP-29747.

src/modules/sub-items/sub.items.module.js
tischsoic: I renamed it to `notMovedLocations` so that it will be similar to `movedLocations` and won't suggest that it is `item === { location: {...}, conent: {...} }`.

tischsoic:
ping @sunpietro 
I applied fixes. I also slightly changed the behaviour - now during bulk operation (bulk move/bulk delete) we add `m-sub-items__list--loading` class to component so that user won't be able to mess up during requests to server.
Changes are in last commit.

sunpietro (approved):

micszo (approved):

PR #108, opened 2018-10-17
EZP-29737: Invalid UDW container height after resizing window

sunpietro:
https://jira.ez.no/browse/EZP-29737


dew326 (approved):

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -989,6 +1027,14 @@ export default class UniversalDiscoveryModule extends Component {
         return <span className="m-ud__location-not-allowed">{locationNotAllowedMessage}</span>;
     }
 
+    setMainContainerRef(ref) {
+        this._refMainContainer = ref;
```
tischsoic: I would go for `React.createRef()`.

tischsoic, line 1026: I would go for `React.createRef()`.

tischsoic: But if we want to stick with this way in order to look it the same as in other places, then OK.

sunpietro: I wanted to keep the consistency. If we decide to use `React.createRef()` here and there, then we should change it everywhere in this repository.

tischsoic (approved):

sunpietro:
@micszo can you take care of it, or assign someone, please? :)

micszo:
@sunpietro ok :) could you buff up the description with some steps?

sunpietro:
@micszo and other QA members: change the browser window height from tall to short and you should notice, that with this fix the buttons at the bottom of UDW don't disappear below the fold anymore.

barbaragr (changes requested):
When I change height, buttons slide down outside the UDW
[udw_btns.zip](https://github.com/ezsystems/ezplatform-admin-ui-modules/files/2491551/udw_btns.zip)

sunpietro:
@barbaragr this is expected behaviour, because the container height is updated with a delay, to avoid unneccesary calculations which are really expensive in terms of performance.

barbaragr (approved):

lserwatka:
@sunpietro you can merge it up

sunpietro:
@lserwatka done :)

PR #112, opened 2018-10-26
EZP-29714: As an Editor, I want to use pagination in the sub-items widgets

tischsoic:
# Pagination for Sub-items
Implementation of pagination for Sub-items component.

## Ticket
JIRA: https://jira.ez.no/browse/EZP-29714

## Edge cases for testing (for me and *QA*)
- when you delete/move last item from the given page it should fall back to previous page (if exists)
- when you delete/move all sub-item it should display information that there are no items
- if elements are sorted by *Priority* and you change priority of some element, all pages data should be discarded and loaded again
- if there are no sub-items switch view button should be disabled

## Screenshots

screen shot 2018-10-31 at 16 11 06

When there are no subitems: 
screen shot 2018-11-06 at 14 54 44

Pagination view with many pages:
screen shot 2018-11-07 at 11 35 51


tischsoic:
Tests are currently failing because we replaced "Load more" with pagination hence the behaviour has changed. Tests need to be adjusted to new pagination.

tischsoic:
ping @sunpietro @dew326 - frontend is ready for review

tischsoic:
ping @sunpietro @dew326 - pagination algorithm changed

src/modules/sub-items/components/pagination/pagination.btn.component.js
```diff
@@ -0,0 +1,38 @@
...
+        }
+    };
+
+    let className = `page-item ${additionalClasses}`;
+
+    className = disabled ? `${className} disabled` : className;
+
+    return (
+        <li className={className}>
+            <button className="page-link" onClick={onClick}>
```
dew326: `onClick` -> `onClick` is for me not proper. Because `on click` you do not do `on click` but rather `handle click`

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,129 @@
...
+            pages.push(pagesNumber);
+        }
+
+        return pages;
+    }
+
+    renderPrevious() {
+        const { currentPage, onPageChange } = this.props;
+        const previousPage = currentPage - 1;
+        const hasPreviousPage = currentPage !== 1;
```
dew326: Maybe just `isDisabled = currentPage === 1`? or you can name it `isFirstPage`.

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,129 @@
...
+
+        return (
+            <PaginationBtnComponent
+                page={previousPage}
+                label="Back"
```
dew326: translation

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,129 @@
...
+                disabled={!hasPreviousPage}
+                onPageChange={onPageChange}
+            />
+        );
+    }
+
+    renderNext() {
+        const { currentPage, onPageChange } = this.props;
+        const nextPage = currentPage + 1;
+        const hasNextPage = currentPage !== this.getPagesNumber();
```
dew326: The same story like in `hasPreviousPage`. `isDisabled` or `isLastPage`, `currentPage === this.getPagesNumber()`

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,129 @@
...
+
+        return (
+            <PaginationBtnComponent
+                page={nextPage}
+                label="Next"
```
dew326: translation

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,129 @@
...
+            />
+        );
+    }
+
+    renderPages(pages) {
+        const { currentPage, onPageChange } = this.props;
+
+        return pages.map((page) => {
+            if (page === '...') {
+                return <PaginationBtnComponent label="..." disabled={true} />;
```
dew326: This component also should have some `key` prop.

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -249,16 +155,16 @@ export default class TableViewComponent extends Component {
                     <td className={cellHeadClass}>
                         <input type="checkbox" checked={anyLocationSelected} onChange={this.onSelectAll} />
                     </td>
-                    <td className={`${cellHeadClass} ${cellClass}--name ${cellSortClass}`} onClick={this.sortByName}>
+                    <td className={`${cellHeadClass} ${cellClass}--name ${cellSortClass}`} onClick={this.onSortByName}>
```
dew326: The previous one was proper. `onClick` do `sortByName` not `onSortByName`.
Also applies to `onSortByDate` and `onSortByPriority`

src/modules/sub-items/components/viewing-message/viewing.message.component.js
```diff
@@ -0,0 +1,30 @@
...
+        {
+            viewingCount,
+            totalCount,
+        },
+        'sub_items'
+    );
+
+    return (
+        <div className="c-viewing-message">
+            <div className="c-viewing-message__message" dangerouslySetInnerHTML={{ __html: message }} />
```
dew326: `c-viewing-message__message` looks not so good. Maybe `c-viewing-message__text`?

src/modules/sub-items/sub.items.module.js
```diff
@@ -605,18 +806,65 @@ export default class SubItemsModule extends Component {
...
+                    contentTypesMap={this.state.contentTypesMap}
+                    handleItemPriorityUpdate={this.handleItemPriorityUpdate}
+                    items={pageItems}
+                    languages={this.props.languages}
+                    handleEditItem={this.props.handleEditItem}
+                    generateLink={this.props.generateLink}
+                    onItemSelect={this.toggleItemSelect}
+                    toggleAllItemsSelect={this.toggleAllPageItemsSelection}
+                    selectedLocationsIds={selectedPageLocationsIds}
+                    onSortChange={this.onSortChange}
```
dew326: `onSortChange ={this.changeSorting}`.

src/modules/sub-items/sub.items.module.js
```diff
@@ -572,23 +750,46 @@ export default class SubItemsModule extends Component {
...
-                totalCount={this.state.totalCount}
-                loadedCount={this.state.items.length}
-                limit={this.state.limit}
-                onLoadMore={this.handleLoadMore}
+            <PaginationComponent
+                proximity={1}
+                itemsPerPage={itemsPerPage}
+                currentPage={activePage + 1}
+                totalCount={totalCount}
+                onPageChange={this.onPageChange}
```
dew326: `onPageChange={this.changePage}`.

src/modules/sub-items/sub.items.module.js
```diff
@@ -60,9 +73,26 @@ export default class SubItemsModule extends Component {
...
     }
 
-    componentDidUpdate(prevProps, prevState) {
-        if (prevState.offset < this.state.offset) {
-            this.loadItems();
+    componentDidUpdate() {
+        const { activePage, pages, totalCount} = this.state;
+        const { limit: itemsPerPage } = this.props;
+        const numberOfPages = Math.ceil(totalCount / itemsPerPage);
+        const pageDoesNotExist = activePage > numberOfPages - 1 && activePage !== 0;
```
dew326: `isPageExisting`?

src/modules/sub-items/sub.items.module.js
```diff
@@ -74,14 +104,34 @@ export default class SubItemsModule extends Component {
...
+        }
+
+        const name = objKeys[0];
+        const order = sortClauses[name];
+
+        return { name, order };
+    }
+
+    updateListViewHeight() {
+        const listViewNode = document.querySelector('.m-sub-items__list');
```
dew326: Maybe `ref`?

src/modules/sub-items/sub.items.module.js
```diff
@@ -90,32 +140,34 @@ export default class SubItemsModule extends Component {
...
-            .then(this.loadContentItems)
+        this.loadLocation(page)
+            .then(this.loadContentItems.bind(this, page))
             .then(this.loadContentTypes)
             .then(this.updateItemsState)
-            .catch(() =>
-                window.eZ.helpers.notification.showErrorNotification('An error occurred while loading items in the Sub Items module')
-            );
+            .catch(() => {
+                window.eZ.helpers.notification.showErrorNotification('An error occurred while loading items in the Sub Items module');
```
dew326: Could you add translation?

src/modules/sub-items/sub.items.module.js
```diff
@@ -196,22 +252,26 @@ export default class SubItemsModule extends Component {
      * @param {Object} responses
      * @memberof SubItemsModule
      */
-    updateItemsState({ locations, content, totalCount, contentTypes }) {
-        const items = locations.reduce((total, location) => {
+    updateItemsState({ page, locations, contents, totalCount, contentTypes }) {
+        const pageItems = locations.map((location) => {
             const itemLocation = location.value.Location;
+            const respectiveContent = contents.find((item) => item.value.Content._id === itemLocation.ContentInfo.Content._id).value
+                .Content;
```
dew326: When this `.Content` is in another line it looks weird. Can we do something with this?

src/modules/sub-items/sub.items.module.js
```diff
@@ -240,6 +300,49 @@ export default class SubItemsModule extends Component {
         }, {});
     }
 
+    updateTotalCountState(totalCount) {
+        this.setState(() => {
```
dew326: Is this working without the brackets? Shouldn't be `({ totalCount })`?

src/modules/sub-items/sub.items.module.js
```diff
@@ -240,6 +300,49 @@ export default class SubItemsModule extends Component {
...
+                [currentPage]: null,
+            },
+        }));
+    }
+
+    onSortChange(sortClause) {
+        this.updateListViewHeight();
+        this.setState((state) => ({
+            sortClause,
+            sortOrder: state.sortClause === sortClause ? (state.sortOrder === 'ascending' ? 'descending' : 'ascending') : 'ascending',
```
dew326: This ternary operator is to complicated.

src/modules/sub-items/sub.items.module.js
```diff
@@ -293,72 +431,91 @@ export default class SubItemsModule extends Component {
...
+     * Page data might be discarded but selection of the item should stay,
+     * thus we need item data to perform operations such as bulk delete or bulk move
+     *
+     * @param {{content: Object, location: Object}} item
+     */
+    getItemDataToCache(item) {
+        const { contentTypesMap } = this.state;
+        const notAvailableLabel = Translator.trans(/*@Desc("N/A")*/ 'content_type.not_available.label', {}, 'sub_items');
+        const contentType = contentTypesMap[item.content.ContentType._href];
+        const contentTypeName = contentType ? contentType.names.value[0]['#text'] : notAvailableLabel;
```
dew326: This will return the name in the first language. The translation will not work.

src/modules/sub-items/sub.items.module.js
```diff
@@ -557,6 +718,23 @@ export default class SubItemsModule extends Component {
         );
     }
 
+    onPageChange(page) {
+        this.updateListViewHeight();
+        this.setState(() => ({
+            activePage: page - 1,
```
dew326: Why activePage is page - 1?

src/modules/sub-items/sub.items.module.js
```diff
@@ -572,23 +750,46 @@ export default class SubItemsModule extends Component {
...
-     * @method renderLoadMore
+     * @method renderViewingMessage
+     * @returns {JSX.Element}
+     */
+    renderViewingMessage() {
+        const { totalCount } = this.state;
+        const { activePage, pages } = this.state;
+        const pageData = pages[activePage];
+        const pageItems = pageData ? pageData.items : [];
+        const viewingCount = pageItems ? pageItems.length : 0;
```
dew326: hmm `pageItems` is an array (empty or populated), so `pageItems.length` is enough?

dew326 (changes requested):

src/modules/sub-items/components/pagination/pagination.btn.component.js
tischsoic: I like this proposition. I was curious to know how it was for the rest of the code in `admin-ui-modules`, so I made small search. For `={han` search there are 2 results, and for `={on` there are 25 results, which have been written by me, you and @sunpietro . 😛 

I will gladly change it to `handleClick` because it feels better but it's hard to write code which satisfies others when there are no clear rules written down. I think I will start writing them down. 😄

lserwatka: @tischsoic It is a very good idea!

dew326: There is so many `={on` because of passing prop from parent to the child component. This is ok, the prop can be named `onSomething`, we should avoid naming method `onSomething`. 

When I checked for `={this.on` there are a few matches (mostly from the beginning of v2 where we were in kinda rush).

tischsoic: Thank you for reply and clarification @dew326 , now it makes sense to me. :)

tischsoic:
I've just realised that we have `activePage` in sub.items.module.js and `currentPage` in `pagination.component.js`. Which name do we want? Or maybe leave it as it is?
ping @sunpietro @dew326

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: Nice catch, interesting that I don't recall any React warnings. 🤔 But it should have a key.

src/modules/sub-items/components/pagination/pagination.btn.component.js
tischsoic: During implementation of bulk delete/move I introduced `onSelectAll` method: https://github.com/ezsystems/ezplatform-admin-ui-modules/blob/946fdedb0895ffb7ed811adcfb0884bdb64adbbc/src/modules/sub-items/components/table-view/table.view.component.js#L154
I'm wondering, should it be named `selectAll` or `handleSelectAll`? ping @dew326 :)

tischsoic: OK, now I saw other comments and I understand it should be renamed to `selectAll`.

src/modules/sub-items/sub.items.module.js
tischsoic: Because in the pagination the page numbers start from 1 and thus pagination button is passing `n`, not `n-1` (e.g 1 instead of 0). 
I was thinking about it and your comment convinced me that it should be changed.

src/modules/sub-items/sub.items.module.js
tischsoic: No 😄 `pageData` contains `{ isLoading: boolean, items: Array }` and items may be null, because in `setPageLoadingState` I set the pageData to: `{ isLoading: true, items: null }`.

I'm setting it to `null`, because I have this condition:
```js
const activePageData = pages[activePage];
const showSpinner = !activePageData || !activePageData.items;

if (showSpinner) {
     return this.renderSpinner();
}
```
in `renderListView`, because when we don't have subitems in given location we want to pass empty array to `SubItemsListComponent` (as before) and `<NoItemsComponent>` is rendered by them. What do you think? Maybe I should change it so that we would render `<NoItemsComponent>` in `sub.items.component.js` based on `state.totalCount`? It may be simpler.

src/modules/sub-items/sub.items.module.js
tischsoic: Hmm, I took this piece of code from somewhere else, we should use `window.eZ.adminUiConfig.contentTypeNames` now?

src/modules/sub-items/sub.items.module.js
tischsoic: I solved it by adding `map` in `loadContentItems`: 
`contents: contentInfo.View.Result.searchHits.searchHit.map((searchHit) => searchHit.value.Content)`.
Now we don't have `.Content` in another line and `contents` contains `Content`s and not `searchHit.value.Content`s.

src/modules/sub-items/sub.items.module.js
dew326: It seems like overcomplicated.

Correct me if I'm wrong:
1. `renderSpinner()` - renders spinner when items are loading.
2. You have flag `isLoading`
3. You render spinner when there are no items

Why not when `isLoading` is `true`?

src/modules/sub-items/sub.items.module.js
dew326: Yes

src/modules/sub-items/sub.items.module.js
tischsoic: > Why not when isLoading is true?

Because on the first hit you don't have `pageData` for given `activePage` - it is set in `componentDidUpdate`, so on the first hit `pageData` is undefined so basically `pageData. isLoading == false` (because it is undefined) but you should treat it as `isLoading`.

dew326: It feels like something is wrong in the architecture, we have flag `isLoading` but the flag doesn't answer the question if page is loading.

Resources/translations/sub_items.en.xliff
```diff
@@ -111,6 +101,21 @@
...
+        <target state="new">An error occurred while loading items in the Sub Items module</target>
+        <note>key: page.loading_error.message</note>
+      </trans-unit>
+      <trans-unit id="b10157de76c0f57fa166904d97c06336e10559ca" resname="pagination.next">
+        <source>Next</source>
+        <target state="new">Next</target>
+        <note>key: pagination.next</note>
+      </trans-unit>
+      <trans-unit id="c6a31daba40a2cc95badf8a9c20773f144f1601b" resname="pagination.previous">
+        <source>Back</source>
```
sunpietro: The identifier says `previous` the label says `back`. Maybe you should consolidate the naming convention?

src/modules/sub-items/components/pagination/pagination.btn.component.js
```diff
@@ -0,0 +1,38 @@
...
+import PropTypes from 'prop-types';
+
+const PaginationBtnComponent = ({ label, disabled, additionalClasses, onPageChange, page }) => {
+    const handleClick = () => {
+        if (!disabled && Number.isInteger(page)) {
+            onPageChange(page);
+        }
+    };
+
+    let className = `page-item ${additionalClasses}`;
```
sunpietro: The CSS class name is incosistent with the convention exisiting in this repo.

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,126 @@
...
+        for (let i = 1; i <= pagesNumber; i++) {
+            const isFirstPage = i === 1;
+            const isLastPage = i === pagesNumber;
+            const isInRange = i >= (currentPage + 1) - proximity && i <= (currentPage + 1) + proximity;
+
+            if (isFirstPage || isLastPage || isInRange) {
+                pages.push(i);
+                wereDots = false;
+            } else if (!wereDots) {
+                pages.push('...');
```
sunpietro: I would convert dots into a kind of static variable that could be reused in several places.

```suggestion
                const DOTS = '...';
                pages.push(DOTS);
```

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,126 @@
...
+            );
+        });
+    }
+
+    render() {
+        const pages = this.computePages();
+
+        return (
+            <div className="c-pagination row justify-content-center align-items-center">
+                <ul className="pagination ez-pagination">
```
sunpietro: Incorrect CSS class name convention

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -18,140 +20,42 @@ export default class TableViewComponent extends Component {
...
-        if (sortKey === SORTKEY_NAME) {
-            this.sortByName(items);
-        } else if (sortKey === SORTKEY_DATE) {
-            this.sortByDate(items);
-        } else if (sortKey === SORTKEY_PRIORITY) {
-            this.sortByPriority(items);
-        } else {
-            this.setState(() => ({ items }));
-        }
+        onSortChange('ContentName');
```
sunpietro: I would make it a constant like `const KEY_CONTENT_NAME = 'ContentName';`. Then in the:

```
const SORTKEY_MAP = {
    [KEY_CONTENT_NAME]: 'name',
    DateModified: 'date',
    LocationPriority: 'priority',
};
```

sunpietro: The similar rule applies to other places.

src/modules/sub-items/components/viewing-message/viewing.message.component.js
```diff
@@ -0,0 +1,30 @@
...
+        </div>
+    );
+};
+
+ViewingMessageComponent.propTypes = {
+    totalCount: PropTypes.number.isRequired,
+    viewingCount: PropTypes.number.isRequired,
+};
+
+ViewingMessageComponent.defaultProps = {};
```
sunpietro: It can be removed.

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -18,140 +20,42 @@ export default class TableViewComponent extends Component {
...
+     * Changes sort to name
      */
-    sortItems(items) {
-        const sortKey = this.state.sortKey;
-
-        if (!sortKey) {
-            this.setState(() => ({ items }));
-        }
+    sortByName() {
+        const { onSortChange } = this.props;
```
sunpietro, line 32: Where is this method/function defined?

sunpietro (changes requested):

Resources/translations/sub_items.en.xliff
tischsoic: I was thinking about it, I will change this.

src/modules/sub-items/components/pagination/pagination.btn.component.js
tischsoic: You mean `page-item`? AFAIK it's Bootstrap class, do you want me to duplicate it?

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: The same as above.

src/modules/sub-items/components/pagination/pagination.btn.component.js
sunpietro: If it's from Bootstrap then it's fine.

src/modules/sub-items/components/table-view/table.view.component.js
tischsoic: in `sub.items.module.js`, it is past via props here: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/112/files#diff-bebc1f75bde44101c137d4bc83684925R872

src/modules/sub-items/components/pagination/pagination.btn.component.js
tischsoic: OK 👍

src/modules/sub-items/sub.items.module.js
```diff
@@ -240,6 +304,57 @@ export default class SubItemsModule extends Component {
...
+                [currentPage]: null,
+            },
+        }));
+    }
+
+    changeSorting(sortClause) {
+        this.updateListViewHeight();
+        this.setState((state) => ({
+            sortClause,
+            sortOrder: state.sortClause === sortClause ? this.getOppositeSortOrder(state.sortOrder) : this.getDefaultSortOrder(),
```
sunpietro: I would expose it as a class method: `getSortOrder`: `state.sortClause === sortClause ? this.getOppositeSortOrder(state.sortOrder) : this.getDefaultSortOrder()`

What do you think about this approach?

tischsoic: This was my first approach but `getSortOrder` is not very descriptive and you have to pass `state.sortClause`, `state.sortOrder` and `sortClause` to this method which for me obscures what it does.
But if you prefer `getSortOrder` I can change it. :)

src/modules/sub-items/components/pagination/pagination.btn.component.js
```diff
@@ -0,0 +1,38 @@
...
+
+    className = disabled ? `${className} disabled` : className;
```
sunpietro: Is the `disabled` class coming from Bootstrap? Why not to add `disabled` attribute on the button?

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,126 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+
+import PaginationBtnComponent from './pagination.btn.component';
+
+export default class PaginationComponent extends PureComponent {
```
sunpietro: It should be a functional component instead.

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,126 @@
...
+                label={previousLabel}
+                additionalClasses="prev"
+                disabled={isFirstPage}
+                onPageChange={onPageChange}
+            />
+        );
+    }
+
+    renderNext() {
+        const { currentPage, onPageChange } = this.props;
```
sunpietro: I would rename `currentPage` to `currentPageIndex`

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,126 @@
...
+    renderPages(pages) {
+        const { currentPage, onPageChange } = this.props;
+
+        return pages.map((page, index) => {
+            if (page === '...') {
+                return <PaginationBtnComponent key={index} label="..." disabled={true} />;
+            }
+
+            const isCurrentPage = page === currentPage + 1;
+            const additionalClasses = isCurrentPage ? 'active' : '';
```
sunpietro: Is `active` CSS class coming from Bootstrap?

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,126 @@
...
+                />
+            );
+        });
+    }
+
+    render() {
+        const pages = this.computePages();
+
+        return (
+            <div className="c-pagination row justify-content-center align-items-center">
```
sunpietro: I bet, the `row` CSS class should not be here.

sunpietro (changes requested):

src/modules/sub-items/sub.items.module.js
sunpietro: We should definitely use this flag to display the loading spinner, unless we want to mess up our code. If the `pageData` is undefined, then you have to find the answer why is it set this way.
On first render we get all data for subitems from the backend, so no initial loading is required. You have to check how the code works and understand it. Adding additional code that duplicates existing solutions is not right way of development.

src/modules/sub-items/sub.items.module.js
sunpietro: I think, `getSortOrder` makes sense.

sunpietro:
> I've just realised that we have `activePage` in sub.items.module.js and `currentPage` in `pagination.component.js`. Which name do we want? Or maybe leave it as it is?

I would stick with `activePage`

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: This is how it is styled in other places, we use `row` css class.

src/modules/sub-items/components/pagination/pagination.btn.component.js
tischsoic: Styling does not work if I switch to attribute.

src/modules/sub-items/components/pagination/pagination.component.js
sunpietro: Shouldn't the `row` class be applied to the nested element instead: `ul`? Because it will be the row actually.

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: Yes, https://getbootstrap.com/docs/4.1/components/pagination/#disabled-and-active-states

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: I decided to use class component because with that many functions I think it is more readable and clear. Do you want me to change it to functional component and change each method to arrow function within it?

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: Mhm, I like this idea, you want to add `Index` to suggest that numeration in `currentPage` variable starts from 0? :)

src/modules/sub-items/components/pagination/pagination.component.js
sunpietro: The functions inside the class can be easily converted to a _standard_ arrow functions outside of functional component. They will take props as arguments and return a value.
I believe it's better not to define such functions inside the functional component as they're not very specific.

src/modules/sub-items/components/pagination/pagination.component.js
sunpietro: That would make sense. And it should be `activePage`, right?

src/modules/sub-items/sub.items.module.js
tischsoic: > If the pageData is undefined, then you have to find the answer why is it set this way.

I think that my explanation wasn't clear enough, when saying this:
> Because on the first hit you don't have pageData for given activePage - it is set in componentDidUpdate, so on the first hit pageData is undefined so basically pageData. isLoading == false (because it is undefined) but you should treat it as isLoading.

by `on the first hit` I was talking about "first hit" after switching to a page not yet loaded - so I don't have data for it. I trigger loading in `componentDidUpdate`. :) 

So it's like:
1. switch to a new page (not loaded yet),
2. render triggers after state change (`pageData` not set),
3. detect in `componentDidUpdate` that activePage has not been loaded yet and trigger loading (set `isLoading` flag).

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: Right 👍

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: It can be moved to `ul`, I will do this.

tischsoic: I moved CSS classes to `ul` element and removed `div`.

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: I changed this but it might be a little bit inconsistent to have same value under `currentPage` and `currentPageIndex.`. 🤔  Nonetheless, I will keep it as it is now, after correction.

tischsoic:
ping @dew326 @sunpietro

src/modules/sub-items/sub.items.module.js
```diff
@@ -293,72 +447,96 @@ export default class SubItemsModule extends Component {
...
+     * @param {Object} item.location
+     *
+     * @returns {Object} dataToCache
+     */
+    getItemDataToCache(item) {
+        const { contentTypesMap } = this.state;
+        const notAvailableLabel = Translator.trans(/*@Desc("N/A")*/ 'content_type.not_available.label', {}, 'sub_items');
+        const contentType = contentTypesMap[item.content.ContentType._href];
+        const contentTypeIdentifier = contentType.identifier;
+        const contentTypeName = window.eZ.adminUiConfig.contentTypeNames[contentTypeIdentifier] || notAvailableLabel;
```
dew326: If you have `contentTypeIdentifier` it will always have the name in the object `contentTypeNames`. I mean `|| notAvailableLabel` will never be shown. 
The only one what you may not have is content type in `contentTypesMap` but it will throw an exception earlier when you try to read property `identifier`.

dew326 (changes requested):

src/modules/sub-items/sub.items.module.js
tischsoic: Fix amended to last commit.

src/modules/sub-items/components/pagination/pagination.btn.component.js
sunpietro: Then I would create a generic selector: `[disabled] { ... }` in `vendor/ezsystems/ezplatform-admin-ui/src/bundle/Resources/public/scss/_general.scss` and I'd add the properties required to display a disabled element the proper way. What do you think about it @dew326 ?

src/modules/sub-items/components/pagination/pagination.btn.component.js
```diff
@@ -0,0 +1,38 @@
...
+
+    return (
+        <li className={className}>
```
sunpietro: I was thinking about this situation when a component uses only Bootstrap CSS classes for styling. I've came up with an idea that it should have at least one CSS class that tell us that this is a `PaginationButtonComponent`. I'd add a CSS class: `c-pagination-button` to the `li` element.

src/modules/sub-items/components/pagination/pagination.btn.component.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const PaginationBtnComponent = ({ label, disabled, additionalClasses, onPageChange, page }) => {
```
sunpietro: `PaginationButtonComponent`

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,101 @@
...
+const PaginationComponent = ({ totalCount, itemsPerPage, proximity, activePageIndex, onPageChange }) => {
+    const pagesNumber = Math.ceil(totalCount / itemsPerPage);
+    const backLabel = Translator.trans(/*@Desc("Back")*/ 'pagination.back', {}, 'sub_items');
+    const nextLabel = Translator.trans(/*@Desc("Next")*/ 'pagination.next', {}, 'sub_items');
+    const previousPage = activePageIndex - 1;
+    const nextPage = activePageIndex + 1;
+    const isFirstPage = activePageIndex === 0;
+    const isLastPage = activePageIndex === pagesNumber - 1;
+    const pages = computePages({ proximity, activePageIndex, pagesNumber });
+    const pagesElements = pages.map((page, index) => {
```
sunpietro: What does `page`variable contain?

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,101 @@
...
+                page={page - 1}
+                label={label}
+                additionalClasses={additionalClasses}
+                onPageChange={onPageChange}
+            />
+        );
+    });
+
+    return (
+        <ul className="pagination row justify-content-center ez-pagination">
```
sunpietro: I'd add a CSS class: `c-pagination`

sunpietro: Do you need `ez-pagination` class?

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,101 @@
...
+ * Example 3: [ 1, "...", 8, 9, 10, 11, 12, "...", 20 ] (for: proximity = 2; pagesNumber = 20; activePageIndex = 10)
+ *
+ * @param {Object} params
+ * @param {Number} params.proximity
+ * @param {Number} params.activePageIndex
+ * @param {Number} params.pagesNumber
+ *
+ * @returns {Array}
+ */
+const computePages = ({ proximity, activePageIndex, pagesNumber }) => {
```
sunpietro: Please, place it before component definition.

src/modules/sub-items/sub.items.module.js
```diff
@@ -60,9 +74,26 @@ export default class SubItemsModule extends Component {
...
-        if (prevState.offset < this.state.offset) {
-            this.loadItems();
+    componentDidUpdate() {
+        const { activePage, pages, totalCount } = this.state;
+        const { limit: itemsPerPage } = this.props;
+        const numberOfPages = Math.ceil(totalCount / itemsPerPage);
+        const pageDoesNotExist = activePage > numberOfPages - 1 && activePage !== 0;
+
+        if (pageDoesNotExist) {
+            this.setState({
```
sunpietro, line 82: This is risky, as it might lead to infinite loop. We have to be really careful when updating state in the lifecycle method that is run right after a state is updated.

src/modules/sub-items/sub.items.module.js
```diff
@@ -141,51 +199,54 @@ export default class SubItemsModule extends Component {
...
-        return Promise.all(promises);
+        return new Promise((resolve) => {
+            const contentIds = locations.map((item) => item.value.Location.ContentInfo.Content._id);
+
+            loadContentInfo(restInfo, contentIds, (contentInfo) => {
+                const { totalCount: oldTotalCount } = this.state;
+                const totalCount = response.View.Result.count;
+
+                if (oldTotalCount !== totalCount) {
+                    this.discardAllPagesOtherThan(page);
```
sunpietro: This is strange approach for naming methods: `discardAllPagesOtherThan()`. I would rename it to `discardPages(<array-of-excluded-pages-from-discard>)`.
What is a result of discard action?

src/modules/sub-items/sub.items.module.js
```diff
@@ -240,6 +304,61 @@ export default class SubItemsModule extends Component {
...
+    }
+
+    discardAllPages() {
+        this.updateListViewHeight();
+        this.setState(() => ({
+            pages: {},
+        }));
+    }
+
+    discardCurrentPage() {
```
sunpietro: `discardCurrentPageData`

src/modules/sub-items/sub.items.module.js
```diff
@@ -240,6 +304,61 @@ export default class SubItemsModule extends Component {
...
+    discardAllPagesOtherThan(page) {
+        this.updateListViewHeight();
+        this.setState((state) => ({
+            pages: {
+                [page]: state.pages[page],
+            },
+        }));
+    }
+
+    discardAllPages() {
```
sunpietro: `discardAllPagesData`

src/modules/sub-items/sub.items.module.js
```diff
@@ -240,6 +304,61 @@ export default class SubItemsModule extends Component {
         }, {});
     }
 
+    updateTotalCountState(totalCount) {
+        this.setState(() => ({
+            totalCount,
+        }));
+    }
+
+    discardAllPagesOtherThan(page) {
```
sunpietro: This is redundant. This case can be covered in `discardAllPagesData` method

src/modules/sub-items/sub.items.module.js
```diff
@@ -259,27 +378,62 @@ export default class SubItemsModule extends Component {
...
      */
-    updateItemsOrder(location, state) {
-        const items = state.items;
-        const item = items.find((element) => element.location.id === location.id);
+    findItemPositionByLocationId(pages, locationId) {
+        let itemPage = null;
+        let itemIndex = null;
 
-        item.location = location;
+        for (const [page, pageData] of Object.entries(pages)) {
```
sunpietro: Why not `Array.find`?

src/modules/sub-items/sub.items.module.js
```diff
@@ -293,72 +447,95 @@ export default class SubItemsModule extends Component {
...
+     * Page data might be discarded but selection of the item should stay,
+     * thus we need item data to perform operations such as bulk delete or bulk move
+     *
+     * @param {Object} item
+     * @param {Object} item.content
+     * @param {Object} item.location
+     *
+     * @returns {Object} dataToCache
+     */
+    getItemDataToCache(item) {
```
sunpietro: I don't get the idea of caching here. Can you elaborate more on this topic? What is the flow?

src/modules/sub-items/sub.items.module.js
sunpietro: Maybe you should set base, empty state for a page without data when trying to load it?

sunpietro (changes requested):

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: One value of the array: `[ 1, "...", 3, 4, 5, 6, 7, "...", 10 ]`.

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: It seems that we can remove `ez-pagination` when it comes to solely styling.

src/modules/sub-items/sub.items.module.js
tischsoic: I will try. :)

src/modules/sub-items/sub.items.module.js
tischsoic: You cache item data during a selection of a page item so that you have item data even if page data were discarded.

tischsoic: When it comes to an example of flow: 
1. you open the site, Sub-items 1st page has been loaded
2. you sort items by name *ascending*
3. 1st page reloads
4. you select 1st item on the 1st page
5. you sort items by name again but *descending*
6. 1st page reloads (now selected item is on another page)
7. you select another 1st item on 1st page
8. you perform bulk delete in which both items should be moved to trash

src/modules/sub-items/sub.items.module.js
tischsoic: I think I have chosen this because `for of` with `break;` seemed for me more clear than `Object.entries(pages).find(callback)` with callback returning `true` after finding given page.

tischsoic: For me both ways are OK. Do you want me to convert this to `Array.find()`?

tischsoic:
ping @sunpietro @dew326 
New changes in last commit.
I applied fixes according to PR review. I added new variable `pagesDuringLoading` which stores set of pages during loading so that pages data in state are only items - no loading information. We show spinner when `pageItems` are `undefined`. I moved `<NoItemsComponent />` higher to simplify logic. +other small changes

tischsoic:
ping @sunpietro @dew326

src/modules/sub-items/components/pagination/pagination.button.component.js
```diff
@@ -0,0 +1,38 @@
...
+            </button>
+        </li>
+    );
+};
+
+PaginationButtonComponent.propTypes = {
+    label: PropTypes.string.isRequired,
+    disabled: PropTypes.bool.isRequired,
+    onPageChange: PropTypes.func,
+    page: PropTypes.number,
```
dew326: `page` seems to be required and `disabled` seems to be not required.

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const PaginationComponent = ({ totalCount, itemsPerPage, proximity, activePageIndex, onPageChange }) => {
+    const pagesNumber = Math.ceil(totalCount / itemsPerPage);
+    const backLabel = Translator.trans(/*@Desc("Back")*/ 'pagination.back', {}, 'sub_items');
+    const nextLabel = Translator.trans(/*@Desc("Next")*/ 'pagination.next', {}, 'sub_items');
+    const previousPage = activePageIndex - 1;
+    const nextPage = activePageIndex + 1;
+    const isFirstPage = activePageIndex === 0;
+    const isLastPage = activePageIndex === pagesNumber - 1;
```
dew326: Maybe better would be `const isLastPage = activePageIndex + 1 === pagesNumber` because `pagesNumber - 1` suggests before last page.

src/modules/sub-items/sub.items.module.js
```diff
@@ -60,9 +73,27 @@ export default class SubItemsModule extends Component {
...
     }
 
-    componentDidUpdate(prevProps, prevState) {
-        if (prevState.offset < this.state.offset) {
-            this.loadItems();
+    componentDidUpdate() {
+        const { activePage, pages, totalCount } = this.state;
+        const { limit: itemsPerPage } = this.props;
+        const numberOfPages = Math.ceil(totalCount / itemsPerPage);
+        const pageDoesNotExist = activePage > numberOfPages - 1 && activePage !== 0;
```
dew326: Why it is possible to have `activePage` bigger than `pages count`?

tischsoic: If you are on the last page and delete all items from the last page this last page disappears (no longer exists). This mechanism switches the active page to the last one.

src/modules/sub-items/components/pagination/pagination.button.component.js
tischsoic: In previous implementation `page` was not required, we can make it required. When it comes to  `disabled` I think it may stay as required - we have it in `defaultProps`.

dew326: It's not about what we can do or not, it's about which props have to be provided to make the component work properly.

To make component work you have to pass the `page` - so this is required
If you will not provide the `disabled` flag the default value will be used - so this is not required.

tischsoic: Oh, now I checked it more carefully, and we have:
```js
if (page === DOTS) {
     return <PaginationButtonComponent key={index} label={DOTS} disabled={true} />;
}
```
so `page` prop cannot be required.

dew326: Ok, but `disabled` is not required.

tischsoic: Correct, I was during fixing this.

src/modules/sub-items/components/pagination/pagination.component.js
sunpietro: `paginationButtons`

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -228,9 +117,11 @@ export default class TableViewComponent extends Component {
             return;
         }
 
-        if (this.state.sortKey) {
-            const headSortClass = this.state.isAscSort ? `${headClass}--sort-asc` : `${headClass}--sort-desc`;
-            const headSortByClass = `${headClass}--sort-by-${this.state.sortKey}`;
+        const { sortClause, sortOrder } = this.props;
+
+        if (sortClause) {
+            const headSortClass = sortOrder === 'ascending' ? `${headClass}--sort-asc` : `${headClass}--sort-desc`;
```
sunpietro: `headSortOrderClass`

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -292,4 +183,9 @@ TableViewComponent.propTypes = {
     onItemSelect: PropTypes.func.isRequired,
     toggleAllItemsSelect: PropTypes.func.isRequired,
     selectedLocationsIds: PropTypes.instanceOf(Set),
+    onSortChange: PropTypes.func.isRequired,
+    sortClause: PropTypes.string.isRequired,
+    sortOrder: PropTypes.string.isRequired,
 };
+
+TableViewComponent.defaultProps = {};
```
sunpietro: Can be removed.

src/modules/sub-items/components/viewing-message/viewing.message.component.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './css/viewing.message.component.css';
+
+const ViewingMessageComponent = ({ totalCount, viewingCount }) => {
```
sunpietro: Isn't `PaginationInfoComponent` name better than this, is it?

src/modules/sub-items/components/pagination/pagination.component.js
```diff
@@ -0,0 +1,101 @@
...
+    const backLabel = Translator.trans(/*@Desc("Back")*/ 'pagination.back', {}, 'sub_items');
+    const nextLabel = Translator.trans(/*@Desc("Next")*/ 'pagination.next', {}, 'sub_items');
+    const previousPage = activePageIndex - 1;
+    const nextPage = activePageIndex + 1;
+    const isFirstPage = activePageIndex === 0;
+    const isLastPage = activePageIndex + 1 === pagesNumber;
+    const pages = computePages({ proximity, activePageIndex, pagesNumber });
+    const pagesElements = pages.map((page, index) => {
+        if (page === DOTS) {
+            return <PaginationButtonComponent key={index} label={DOTS} disabled={true} />;
```
dew326: Using `index` and `page` as key may cause conflicts.
In case: `[ 1, "...", 3, 4, 5, 6, 7, "...", 10 ]`
First iteration: `page === 1`
Second iteration: `index === 1`
In two items key is `1`

Maybe we should use a combination of page and index? like `page-index`?

src/modules/sub-items/sub.items.module.js
```diff
@@ -21,35 +25,44 @@ export default class SubItemsModule extends Component {
...
         this.closeBulkDeletePopup = this.closeBulkDeletePopup.bind(this);
         this.onBulkDeletePopupConfirm = this.onBulkDeletePopupConfirm.bind(this);
         this.afterBulkDelete = this.afterBulkDelete.bind(this);
+        this.changePage = this.changePage.bind(this);
+        this.changeSorting = this.changeSorting.bind(this);
 
+        this._refListViewWrapper = React.createRef();
         this.bulkDeleteModalContainer = null;
         this.udwContainer = null;
+        this.pagesDuringLoading = new Set();
```
sunpietro: This property names still looks enigmatic to me. After getting back to your code, for review, I don't know what is it for.

src/modules/sub-items/sub.items.module.js
```diff
@@ -21,35 +25,44 @@ export default class SubItemsModule extends Component {
...
+
+        const sortClauseData = this.extractFirstSortClauseData(props.sortClauses);
 
         this.state = {
             activeView: props.activeView,
-            items: props.items,
-            selectedLocationsIds: new Set(),
+            pages: {
+                '0': props.items,
```
sunpietro: `0: props.items` looks good as well.

src/modules/sub-items/sub.items.module.js
```diff
@@ -21,35 +25,44 @@ export default class SubItemsModule extends Component {
...
+            },
+            selectedItemsCachedData: new Map(),
```
sunpietro: I still believe that caching is not needed and in practice it's something different than the variable name says to us.

src/modules/sub-items/sub.items.module.js
```diff
@@ -60,9 +73,27 @@ export default class SubItemsModule extends Component {
         document.body.appendChild(this.bulkDeleteModalContainer);
     }
 
-    componentDidUpdate(prevProps, prevState) {
-        if (prevState.offset < this.state.offset) {
-            this.loadItems();
+    componentDidUpdate() {
+        const { activePage, pages, totalCount } = this.state;
+        const { limit: itemsPerPage } = this.props;
+        const numberOfPages = Math.ceil(totalCount / itemsPerPage);
```
sunpietro: `pagesCount`

src/modules/sub-items/sub.items.module.js
```diff
@@ -74,14 +105,23 @@ export default class SubItemsModule extends Component {
...
 
-    /**
-     * Increases an amount of items to be loaded in the list
-     *
-     * @method handleLoadMore
-     * @memberof SubItemsModule
-     */
-    handleLoadMore() {
-        this.setState((state) => ({ offset: state.offset + this.props.limit }));
+    extractFirstSortClauseData(sortClauses) {
```
sunpietro: Maybe `getDefaultSortClause`?

src/modules/sub-items/sub.items.module.js
```diff
@@ -90,32 +130,40 @@ export default class SubItemsModule extends Component {
...
+                );
+
+                window.eZ.helpers.notification.showErrorNotification(errorMessage);
+            });
     }
 
     /**
      * Loads location data
      *
+     * @param {Number} page
```
sunpietro: Maybe `pageNumber` or `pageIndex`?

src/modules/sub-items/sub.items.module.js
```diff
@@ -124,11 +172,12 @@ export default class SubItemsModule extends Component {
      * Loads content items from location
      *
      * @method loadContentItems
+     * @param {Number} page
```
sunpietro: Maybe `pageNumber` or `pageIndex`?

src/modules/sub-items/sub.items.module.js
```diff
@@ -141,51 +190,54 @@ export default class SubItemsModule extends Component {
...
+
+                if (oldTotalCount !== totalCount) {
+                    this.discardPagesDataOtherThan([page]);
+                }
+
+                resolve({
+                    page,
+                    locations,
+                    totalCount,
+                    contents: contentInfo.View.Result.searchHits.searchHit.map((searchHit) => searchHit.value.Content),
```
sunpietro: `content` is singular only, hence you should name it `contentItems`

src/modules/sub-items/sub.items.module.js
```diff
@@ -240,6 +297,42 @@ export default class SubItemsModule extends Component {
         }, {});
     }
 
+    updateTotalCountState(totalCount) {
+        this.setState(() => ({
+            totalCount,
+        }));
+    }
+
+    discardPagesDataOtherThan(pages) {
```
sunpietro: When is this method invoked? Because the method name is a bit messy.

src/modules/sub-items/sub.items.module.js
```diff
@@ -240,6 +297,42 @@ export default class SubItemsModule extends Component {
...
+    getSortOrder(sortClause, newSortClause, currentSortOrder) {
+        return newSortClause === sortClause ? this.getOppositeSortOrder(currentSortOrder) : this.getDefaultSortOrder();
+    }
+
+    getOppositeSortOrder(sortOrder) {
+        return sortOrder === 'ascending' ? 'descending' : 'ascending';
+    }
+
+    getDefaultSortOrder() {
+        return 'ascending';
```
sunpietro: Why not setting it as a const? `const DEFAULT_SORT_ORDER = 'ascending';`

src/modules/sub-items/sub.items.module.js
```diff
@@ -240,6 +297,42 @@ export default class SubItemsModule extends Component {
...
+            pages: {},
+        }));
+    }
+
+    getSortOrder(sortClause, newSortClause, currentSortOrder) {
+        return newSortClause === sortClause ? this.getOppositeSortOrder(currentSortOrder) : this.getDefaultSortOrder();
+    }
+
+    getOppositeSortOrder(sortOrder) {
+        return sortOrder === 'ascending' ? 'descending' : 'ascending';
```
sunpietro: `ascending` and `descending` can be set as constant variables outside the class definition.

src/modules/sub-items/sub.items.module.js
```diff
@@ -259,27 +352,57 @@ export default class SubItemsModule extends Component {
...
-     * Updates items order
+     * Finds on what page item is located
+     * and what is the index of the item in page's items list
      *
-     * @method updateItemsOrder
-     * @param {Object} location
-     * @param {Object} state
-     * @returns {Object}
-     * @memberof SubItemsModule
+     * @returns {{itemPage: number, itemIndex: number}}
```
sunpietro: `pageNumber` and `pageIndex`, maybe?

src/modules/sub-items/sub.items.module.js
sunpietro: Yes, please

src/modules/sub-items/sub.items.module.js
sunpietro: So technically it's not a cache. It's selected items store.
When I see cache, then it means that you can use it get full component state without a need to send requests to backend.

src/modules/sub-items/sub.items.module.js
```diff
@@ -293,72 +416,95 @@ export default class SubItemsModule extends Component {
...
+    getItemDataToCache(item) {
+        const { contentTypesMap } = this.state;
+        const contentType = contentTypesMap[item.content.ContentType._href];
+        const contentTypeIdentifier = contentType.identifier;
+        const contentTypeName = window.eZ.adminUiConfig.contentTypeNames[contentTypeIdentifier];
+
+        return {
+            contentTypeName,
+            name: item.content.Name,
+            location: {
```
sunpietro: Why not storing the full `item` object instead?

src/modules/sub-items/sub.items.module.js
```diff
@@ -293,72 +416,95 @@ export default class SubItemsModule extends Component {
...
+    toggleAllPageItemsSelection(select) {
+        const { pages, activePage } = this.state;
+        const pageItems = pages[activePage];
 
-    selectAllItems() {
-        const { items } = this.state;
-        const contentItemsIds = items.map((item) => item.location.id);
-        const contentItemsIdsSet = new Set(contentItemsIds);
+        if (select) {
+            const itemsCachedData = new Map(pageItems.map((item) => [item.location.id, this.getItemDataToCache(item)]));
```
sunpietro: Do you really need to create a `Map` instance?

src/modules/sub-items/sub.items.module.js
```diff
@@ -293,72 +416,95 @@ export default class SubItemsModule extends Component {
...
-            };
-        });
+        this.setSelectedItemsState(newSelection);
+    }
+
+    deselectAllItems() {
+        this.setSelectedItemsState(new Map());
+    }
+
+    setSelectedItemsState(selectedItemsCachedData) {
```
sunpietro: Why not calling `setState` directly. It's also one function with one param.

src/modules/sub-items/sub.items.module.js
```diff
@@ -605,20 +794,81 @@ export default class SubItemsModule extends Component {
...
+                    <Icon name="spinner" extraClasses="m-sub-items__list-spinner ez-spin ez-icon-x2 ez-icon-spinner" />
+                </div>
+            </div>
+        );
+    }
+
+    renderNoItems() {
+        const { totalCount } = this.state;
+
+        if (totalCount) {
```
sunpietro: `if (this.state.totalCount) {` is just enough

sunpietro (changes requested):
The concept of cache for selected items doesn't look very well. Especially, that it doesn't work like cache. It's more like a store.
Other things are minor compared to the cache.

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: Nice catch! 👍 
When we have dots, under page variable we have: `...`. I think we should be fine with a key like: `"dots-${index}"`.

src/modules/sub-items/sub.items.module.js
tischsoic: I prefer `extractFirstSortClauseData` over `getDefaultSortClause` because although maybe it is a little bit uglier, I know what it does inside. If you remove context from it, function name still makes sense. This is how I reason about this.

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: I'm not sure, you want to rename `pagesElements` to `paginationButtons` or `page` to `paginationButtons`?

sunpietro: I'm sorry for not being clear enough here. I thought about converting `pagesElements` to `paginationButtons` becuase that's what this variable contains.

src/modules/sub-items/components/viewing-message/viewing.message.component.js
tischsoic: It may be. 😝

src/modules/sub-items/sub.items.module.js
sunpietro: I prefer naming methods/functions in the way that I know what it does and what's its purpose. You need this method to get the sort clause when initializing the component, hence why I suggested the name `getDefaultSortClause`. Because when you're initializing something, you'll probably want something to have a default value, like in your code.
If the method was used in many different places, in many different contexts then your reasoning might be correct. Otherwise I prefer my proposal.

dew326: I agree with @sunpietro. The method returns the default sort clause.
If for some reason in initial state you would need to have last sort clause, you would like to introduce the new method `extractLastSortClause`?

src/modules/sub-items/components/pagination/pagination.component.js
tischsoic: Right, it makes sense! :)

src/modules/sub-items/sub.items.module.js
tischsoic: For me both trains of thought are OK, and since we have 2:1 I will change this to `getDefaultSortClause`. :)

tischsoic:
I applied corrections - ping @sunpietro  @dew326

src/modules/sub-items/components/pagination-info/pagination.info.component.js
```diff
@@ -0,0 +1,28 @@
...
+        /*@Desc("Viewing <strong>%viewingCount%</strong> out of <strong>%totalCount%</strong> sub-items")*/ 'viewing_message',
+        {
+            viewingCount,
+            totalCount,
+        },
+        'sub_items'
+    );
+
+    return (
+        <div className="c-pagination-info">
```
sunpietro: I think this `div` is not needed.

src/modules/sub-items/components/pagination-info/pagination.info.component.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './css/pagination.info.component.css';
+
+const PaginationInfoComponent = ({ totalCount, viewingCount }) => {
```
sunpietro: I'm not sure we even need this component as this will be only one `div` if you follow the comment below.

src/modules/sub-items/sub.items.module.js
```diff
@@ -259,27 +323,34 @@ export default class SubItemsModule extends Component {
...
-     * @param {Object} location
-     * @param {Object} state
-     * @returns {Object}
-     * @memberof SubItemsModule
-     */
-    updateItemsOrder(location, state) {
-        const items = state.items;
-        const item = items.find((element) => element.location.id === location.id);
+    updateItemLocation(location) {
+        this.setState((state) => {
```
sunpietro: `this.setState(({ activePageItems }) => {`

src/modules/sub-items/sub.items.module.js
```diff
@@ -605,20 +708,78 @@ export default class SubItemsModule extends Component {
...
+
+        const { listViewHeight } = this.state;
+        const spinnerMinHeight = 90;
+        const style = {
+            height: listViewHeight && listViewHeight > spinnerMinHeight ? listViewHeight : spinnerMinHeight,
+        };
+
+        return (
+            <div style={style}>
+                <div className="m-sub-items__list-spinner-wrapper">
```
sunpietro: `m-sub-items__spinner-wrapper` maybe?

src/modules/sub-items/sub.items.module.js
```diff
@@ -605,20 +708,78 @@ export default class SubItemsModule extends Component {
...
+                    <Icon name="spinner" extraClasses="m-sub-items__list-spinner ez-spin ez-icon-x2 ez-icon-spinner" />
```
sunpietro: `m-sub-items__spinner`

sunpietro (changes requested):

src/modules/sub-items/components/pagination-info/pagination.info.component.js
tischsoic: We can make it with one `div` - I will delete this component and move code to sub.items.module.js.

src/modules/sub-items/sub.items.module.js
tischsoic: `spinner-wrapper` is better 👍

sunpietro (approved):
Seems to be it. Let's see what @dew326 says about it

dew326 (approved):

SylvainGuittard:
🎉 🎉 🎉 
Congrats @tischsoic and thanks @sunpietro @dew326 for the reviews!

barbaragr (approved):
Works fine 

🐴

mnocon:
When merging: please merge https://github.com/ezsystems/ezplatform-admin-ui/pull/710 as well (it adjusts tests)

PR #115, opened 2018-11-28
EZP-29838: Cannot use Multi-file upload on Microsoft Edge

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-29838

Fixes multi-file upload not working on MS Edge.

The error is caused by spread operator, because `list.files` is not an array but rather `FileList`. Problem can be solved by using `Array.from(list.files)` instead of `[...list.files]`.


dew326 (approved):

src/modules/multi-file-upload/multi.file.upload.module.js
```diff
@@ -143,14 +143,14 @@ export default class MultiFileUploadModule extends Component {
...
+     * @param {Event} event
      * @returns {Array}
      * @memberof MultiFileUploadModule
      */
-    proccessUploadedFiles(event) {
+    processUploadedFiles(event) {
         const list = this.extractDroppedFilesList(event);
 
-        return [...list.files].map((file) => ({
+        return Array.from(list.files).map((file) => ({
```
sunpietro, line 153: Wouldn't it be better to use `[].slice.call(list.files).map()` instead?

sunpietro (approved):

src/modules/multi-file-upload/multi.file.upload.module.js
tischsoic: Are you thinking about browser compatibility or something else?
I checked MDN and `Array.from` seems to be supported on all modern browsers (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Browser_compatibility) so in my opinion it's OK to use it here. 
What do you think?

sunpietro: I'm thinking about perfomance

tischsoic: I don't think that we need to optimise this one line. 😄 
And, for me, this notation: `Array.from(list.files)` is better than this: `[].slice.call(list.files)` when it comes to code readability. :)

sunpietro: I'm fine with that

micszo (approved):

lserwatka:
@tischsoic please merge it up

tischsoic:
Merged up to master: https://github.com/ezsystems/ezplatform-admin-ui-modules/commit/b5fa167d7de6c63c3f969e1a23ad41177806d92e

PR #118, opened 2018-12-12
EZP-29873: "Cannot set property 'CurrentVersion' of null" notification when user click Bookmarks tab in UDW

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-29873

## 1 problem

How problem worked under the hood:
1. A user selects some location in _Browse_ or _Create_ tab
2. `contentMeta` is set to some Object
3. in `componentDidUpdate` we start fetching `ContentInfo`
4. we go to _Bookmarks_ tab thus unselecting location - this sets `contentMeta` to `null`
5. _fetch_ from 3rd step is resolved and `updateContentMetaWithCurrentVersion` is called
6.  `updateContentMetaWithCurrentVersion` tries to update `CurrentVersion` on `contentMeta`, which is `null`, because in step 4rd it was unset.

## 2 problem

During fixing the 1st problem I noticed that we are sending more than one request fetching _ContentInfo_. Fix to the 2nd problem is small so I decided to include it in this PR. 

How 2nd problem worked under the hood:
1. user selects some location in _Browse_ or _Create_ tab
2. `contentMeta` is set to some Object and `isPreviewMetaReady` is set to `false`
3. in `componentDidUpdate` we start fetching `ContentInfo`
4. something triggers re-rendering
5. in `componentDidUpdate` _again_ we start fetching `ContentInfo` (because `!!contentMeta && !isPreviewMetaReady` resolves to `true`)
6. response to request from 3rd step arrives - we update `CurrentVersion` and set `isPreviewMetaReady` to `true`
7. 6th step triggers re-render but`!!contentMeta && !isPreviewMetaReady` condition no longer resolve to _true_ because `isPreviewMetaReady` is now `true`


src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -193,15 +195,22 @@ export default class UniversalDiscoveryModule extends Component {
...
-    updateContentMetaWithCurrentVersion(response) {
-        const contentMeta = deepClone(this.state.contentMeta);
+    updateContentMetaWithCurrentVersion(contentId, response) {
+        const { contentMeta } = this.state;
+
+        if (!contentMeta || contentMeta.ContentInfo.Content._id !== contentId) {
+            return;
+        }
+
+        const contentMetaCloned = deepClone(contentMeta);
```
sunpietro: Maybe just `newContentMeta`?

dew326 (approved):

sunpietro (approved):

src/modules/universal-discovery/universal.discovery.module.js
tischsoic: _Cloned_ is not beautiful for me but _new_ neither and _new_ suggests that the whole _contentMeta_ is changing, not only CurrentVersion. But maybe not, I don't know. What do you think, @dew326 ? :)

dew326: So maybe `updatedContentMeta`?

tischsoic: Changed to `updatedContentMeta`. :)

katarzynazawada (approved):

lserwatka:
Please merge it up

tischsoic:
Merged up: https://github.com/ezsystems/ezplatform-admin-ui-modules/commit/5408ea3ee9ab8b233944d62e95baf7713909dd18

PR #120, opened 2018-12-17
EZP-29888: Content Meta Preview appears on Bookmarks tab in UDW

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-29888

## 1st problem

How problem worked under the hood:
1. user opens _Finder_ in _UDW_
2. then, user selects some location, what triggers loading (request to a server is fired), but we are not selecting this location in _UDW_ yet, only in _Finder_
3. user switches to _Bookmarks_ tab
4. response from 2nd point arrives and now we are selecting the location (for which request was called) in _UDW_, but the item is no longer selected in _Finder_ (we are in _Bookmarks_)

Fixed by calling: `this.props.onItemSelect(location);` as soon as user selects location.

## 2nd problem

During fixing the 1st problem I noticed that there is also a problem when you selecting items too fast. (to reproduce I suggest setting slower internet connection) (I also attached recoding on JIRA.)

How 2nd problem worked under the hood:
1. user selects a container in _Finder_ component - let us call this location _A_
2. user clicks on another item (not loaded yet) - it starts loading
3. you click again on container _A_ - it opens immediately as it was already loaded
4. request from 1st point arrives and it overrides selection from the 2nd point, but only partially - only content of the container _A_ is changing, selection stay on _A_ as it should.

~Fixed by checking during processing response if it is the request for the last selection.~
Selection was decoupled from fetching data - we set active elements immediately in `activeLocations` and store data in separate structure: `locationsMap`.


src/modules/universal-discovery/components/finder/finder.component.js
```diff
@@ -300,6 +301,8 @@ export default class FinderComponent extends Component {
 
         this.setState(() => ({ lastSelectedItem: parent }));
 
+        this.lastSelectedLocationId = location.id;
```
dew326: I don't like the name. It will be confusing.
We have: 
`lastSelectedItem` - and this is parent
`lastSelectedLocationId` and this is the id of current selection?

tischsoic:
I fixed another 3 small issues and this PR turned out to be small refactoring.

Other 3 issues were:
1. It seems that "Load more" didn't work for root location (ID: 1) because for root location `location` in `locationsMap` is _null_. To fix this problem I'm checking in `onLoadMore` if parentLocation is _null_ - if it is, I assume it's root location:
`const parentLocationId = parentLocation ? parentLocation.id : ROOT_LOCATION_ID;`
To tackle the root location's problems I introduced two constants: 
`const ROOT_LOCATION_ID = 1;` and `const ROOT_LOCATION_OBJECT = null;`.

2. In _findLocationChildren_ we were fetching children even if `location.childCount` was equal to 0. For me it seemed unnecessary, thus I'm checking if a location has any child and I'm fetching children only if it has any.

3. Before changes: after selecting another container (B), children of previously selected container (A) were displayed till elements of container B were loaded.
Now: after selecting container _B_ we hide children of container _A_ and show children of the _B_ after they are loaded. As discussed with @sunpietro current behaviour should be the correct one. 

Ad 2. But maybe we want to fetch children even if `location.childCount === 0`?

When it comes to refactoring, we no longer manage active locations selection in _FinderTreeBranchComponent_ but in _FinderComponent_ instead. Additionally, `activeLocations` is no longer object, but the array, which stores only location object.
These changes simplified _FinderTreeBranchComponent_ and some methods in _FinderComponent_ like:`updateActiveLocations` or `updateBranchActiveLocations` (which is not needed any more).

I also removed `parent` property from `locationsMap` items because for me was it was misleading and needless as it was containing not object but ID, and it was the same ID as in `location.id`, so `item.parent === item.location.id` and this ID was also a item's key in `locationsMap`.

I would like to change a few other things, such as removing `limit` from _state_ or storing just `item.children = [<Location>]`  instead of `item.data[].value.<Location>` in `locationsMap`, but I didn't want to change too much.

I still need to test if everything is OK, when `isForcedLocation` or `isPreselectedLocation`. But it seems to be OK. 

What do you think, @sunpietro, @dew326 ? Maybe we don't want that big refactoring or maybe some listed issues are actually intended behaviour? 🙂

sunpietro:
@tischsoic can you rebase?

lserwatka:
Would be good to finish this.

tischsoic:
rebased - ping @sunpietro

barbaragr:
It seems to fix this bug partially: https://jira.ez.no/browse/EZP-29490 console is clear, but Load more button still doesn't work.

lserwatka:
@tischsoic are you able to finish this bug? Rebase is needed.

tischsoic:
@barbaragr You can recheck - last commit should have fixed this bug completely I think. :)

barbaragr (changes requested):
Content Items are displayed only in one column in UDW:
http://recordit.co/nQFRE8StRY

tischsoic:
I deleted one CSS import by mistake during rebase, fixed in last commit. ping @barbaragr

barbaragr (approved):
Looks fine

lserwatka:
Could you merge it up?

tischsoic:
Merged-up:
- 1.4: https://github.com/ezsystems/ezplatform-admin-ui-modules/commit/4abb82841862c3425bb260d601d18335525c6f7c
- master: https://github.com/ezsystems/ezplatform-admin-ui-modules/commit/2d4b43fc5ea76491e188c558e9dc6fb08cd47d09

Sorry @lserwatka for late merge.

PR #121, opened 2018-12-19
[WIP] EZP-29937: Content type names are not translated in the Create tab of UDW

sunpietro:
https://jira.ez.no/browse/EZP-29937


sunpietro:
It will require some backend done by @alongosz

dew326:
It seems like very complicated. Getting the preferred language, checking names in all languages, selecting the one from preferred.

In the AdminUiConfig we have the object with `contentType.identifier` and `contentType.name`, why this name cannot be translated by the backend? @alongosz

dew326 (changes requested):
.

sunpietro:
@dew326 I didn't want to mess up with property values as they are used in other contexts

dew326:
Exactly, and all usage of this content type name should be translated.

lserwatka:
@dew326 is right, we should reuse this, or add a new key to this array with translated names.

dew326:
We do not need a new key. We want to all usage of `name` to be translated automatically.

alongosz:
Sorry I didn't see this specific config before. Indeed that seems like the cleanest way. AdminUI PR will be posted in a few minutes. Change of `eZ.adminUiConfig.contentTypeNames` is already working for me there.

sunpietro:
For displaying the content type items' names we're using `contentTypes` prop with the value of `window.eZ.adminUiConfig.contentTypes`. It's passed to the `UniversalDiscoveryModule` when initializing.

dew326:
I think we should change this in both `eZ.adminUiConfig.contentTypeNames` and `eZ.adminUiConfig.contentTypes`.

sunpietro:
Ok, looks like this PR is irrelevant as no changes are required on the frontend side. Right @dew326 and @alongosz ?

alongosz:
Yes, it can be closed.

andrerom:
> I think we should change this in both eZ.adminUiConfig.contentTypeNames and eZ.adminUiConfig.contentTypes

@dew326 be aware these could be combined to one to avoid loading all content types twice on every request

dew326:
@andrerom Yes, I know. We keep it in two because of the BC.

andrerom:
@dew326 well we could find a way to set both js params based on same source data to avoid the extra load, or?

sunpietro:
@andrerom are you aware this PR is closed?

andrerom:
@sunpietro Fully, just like to taunt you ;)

But ok, I'll move the discussion to Slack ;)

dew326:
@andrerom probably, we have to ask backend guys as this config is set there.

PR #124, opened 2018-12-21
EZP-29968: Error 500 for bulk delete/move

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-29968


src/modules/sub-items/services/bulk.service.js
```diff
@@ -54,7 +55,7 @@ const processBulkResponse = (locations, successCode, callback, response) => {
 };
 
 const makeBulkRequest = ({ token }, requestBodyOperations, callback) => {
-    const request = new Request(Routing.generate('ezplatform.bulk_operation'), {
+    const request = new Request(ENDPOINT_BULK, {
         method: 'POST',
         headers: {
             ...HEADERS_BULK,
```
dew326, line 61: I think in this case we should send `X-Siteaccess` in header.

lserwatka: Yes, this is important.

tischsoic: Good point!

dew326 (approved):

ViniTou (approved):

micszo (approved):

PR #125, opened 2019-01-08
EZEE-2660: 500 error occurs when publishing form in form on the fly

sunpietro:
https://jira.ez.no/browse/EZEE-2660

## For QA
When the Form Builder app is opened inside the COTF modal, then the **Publish** button is disabled until an editor closes the Form Builder app.

**Requires:** https://github.com/ezsystems/ezplatform-form-builder/pull/129


src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -140,6 +172,9 @@ ContentCreatorComponent.propTypes = {
     handlePublish: PropTypes.func.isRequired,
     loadLocation: PropTypes.func,
     restInfo: PropTypes.object.isRequired,
+    enablePublishButton: PropTypes.func.isRequired,
```
tischsoic: Are we passing this function (and other 2 below) via props? It seems to be unused inside component.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -37,13 +46,24 @@ export default class ContentCreatorComponent extends Component {
...
+            iframeWindow.onunload = () => {
                 this.setState((state) => Object.assign({}, state, { iframeLoading: true }));
             };
         }
+
+        iframeDoc.body.addEventListener('fbFormBuilderLoaded', this.disablePublishButton, false);
+        iframeDoc.body.addEventListener('fbFormBuilderUnloaded', this.enablePublishButton, false);
+    }
+
+    enablePublishButton() {
```
tischsoic: Maybe we could merge `enablePublishButton` with `disablePublishButton` into `togglePublishButton(isDisabled)`? This might be cleaner. :)

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -11,23 +11,32 @@ export default class ContentCreatorComponent extends Component {
...
+        this.enablePublishButton = this.enablePublishButton.bind(this);
+        this.disablePublishButton = this.disablePublishButton.bind(this);
+        this.renderPublishBtn = this.renderPublishBtn.bind(this);
+
+        this._refIframe = null;
+        this._refContentContainer = null;
 
         this.state = {
             iframeLoading: true,
+            publishBtnDisabled: false,
```
tischsoic, line 22: AFAIK in our code we prefer boolean variables prefixed with _is_, so here we would have `isPublishBtnDisabled`. I drew this conclusion after my PRs reviews, I might be wrong. 😛

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -11,23 +11,32 @@ export default class ContentCreatorComponent extends Component {
 
         this.handleIframeLoad = this.handleIframeLoad.bind(this);
         this.handlePublish = this.handlePublish.bind(this);
+        this.enablePublishButton = this.enablePublishButton.bind(this);
+        this.disablePublishButton = this.disablePublishButton.bind(this);
+        this.renderPublishBtn = this.renderPublishBtn.bind(this);
```
tischsoic, line 16: There is no need to bind it to `this`, as far as I can see.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -84,6 +104,21 @@ export default class ContentCreatorComponent extends Component {
...
 
+    renderPublishBtn() {
+        const publishLabel = Translator.trans(/*@Desc("Publish")*/ 'content_on_the_fly.publish.label', {}, 'universal_discovery_widget');
+        const attrs = {
+            className: 'm-ud__action--publish',
+            onClick: this.handlePublish,
+            type: 'button',
+        };
+
+        if (this.state.publishBtnDisabled) {
```
tischsoic, line 114: For me, cleaner would be to add `disabled: this.state.publishBtnDisabled` to `attrs` above and remove if statement.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -11,23 +11,32 @@ export default class ContentCreatorComponent extends Component {
 
         this.handleIframeLoad = this.handleIframeLoad.bind(this);
         this.handlePublish = this.handlePublish.bind(this);
+        this.enablePublishButton = this.enablePublishButton.bind(this);
+        this.disablePublishButton = this.disablePublishButton.bind(this);
+        this.renderPublishBtn = this.renderPublishBtn.bind(this);
+
+        this._refIframe = null;
+        this._refContentContainer = null;
```
tischsoic: Actually, it seems to be unused.

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -1062,6 +1062,7 @@ export default class UniversalDiscoveryModule extends Component {
                     onCancel={this.props.onCancel}
                     handlePublish={this.handlePublish}
                     restInfo={this.props.restInfo}
+                    setMainContainerRef={this.setMainContainerRef}
```
tischsoic, line 1065: I realised, after analysing _ContentCreatorComponent_, that we have here `setMainContainerRef={` and not `ref={`. It was confusing at first. Maybe cleaner would be to wrap _ContentCreatorComponent_ into forwardRef? But I'm not 100% sure. 🤔

tischsoic (changes requested):

src/modules/universal-discovery/components/content-creator/content.creator.component.js
sunpietro: I made it on purpose. I wanted to have a clear path of function calls without a need to look at function/method params. Simpler for debugging in this case, when we are observing async actions caused by events.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
sunpietro: See the line above. I just keep the convention from that file. In new file, we should name the `boolean` props/attrs/variables with `is` prefix.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
sunpietro: No, I want to set `attrs.disabled` when needed. I don't want to pass `attrs.disabled` with `undefined` value.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
sunpietro: Good catch.

src/modules/universal-discovery/universal.discovery.module.js
sunpietro: I didn't want to rewrite many files to follow the `forwardRef` convention. But, you're right in other cases I would go with `forwardRef` as it's a better approach. Here, I wanted to make it simple without to much rewriting.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
sunpietro: Good catch.

dew326 (approved):

src/modules/universal-discovery/universal.discovery.module.js
tischsoic: Later I also realised that, actually, we need two _refs_ to be set - we are also missing `setContentContainerRef`. So, we even cannot use forwardRef as it only works for one reference.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
tischsoic: With approach that I suggested `attrs.disabled` would never be `undefined`.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
tischsoic: I think we should remove this. 🙂

tischsoic (approved):

katarzynazawada (dismissed):

sunpietro:
@lserwatka the base branch has just been changed

lserwatka:
You can merge it up.

sunpietro:
Done

PR #127, opened 2019-01-09
[RFC] Introduce TypeScript

tischsoic:
#### Proposal to convert our JS code to TypeScript.

There are some different options to adapt old code which we can choose from. I will try to explain them in comments to this PR.

_Currently project does not compile - I didn't want to adapt every file just to show possibilities._ 🙂


src/modules/universal-discovery/components/content-meta-preview/content.meta.preview.component.tsx
```diff
@@ -1,9 +1,29 @@
-import React, { Component, Fragment } from 'react';
-import PropTypes from 'prop-types';
+import * as React from 'react';
+import { Component, Fragment } from 'react';
```
tischsoic, line 2: Imports for React library in TS work a little bit differently. 
`import React, { Component, Fragment } from 'react';` generates error. 

There are some discussions on the Internet, e.g.: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/5128. Nonetheless it seems that the TS import is correct, and the current one from JS files is possible because of how Babel works. Maybe we can force old import by some tsconfig.json options.

But, AFAIK, most React projects just `import * as React from 'react';` and then use `React.Component`, `React.Fragment` etc.

src/modules/universal-discovery/components/content-meta-preview/content.meta.preview.component.tsx
```diff
@@ -1,9 +1,29 @@
...
 
 import './css/content.meta.preview.component.css';
 
-export default class ContentMetaPreviewComponent extends Component {
+type ContentMetaPreviewComponentProps = {
```
tischsoic, line 6: There are 3 possibilities when it comes to converting propTypes:
1) We can convert `MyComponent.propTypes` to type passed to `Component<Props, State>`. This gives compile time safety + enables prompting properties by the editor during writing code.
2) We can leave old prop types and just pass `any` to `Component`: `Component<any, any>` - no checking by TS, no checking during coding in IDE, but still we have dynamic checking during use of the application. (as in e.g. `content.creator.component.tsx` in this PR)
3) Both above options - leave propTypes and add TS types

src/modules/universal-discovery/universal.discovery.module.tsx
```diff
@@ -223,7 +226,7 @@ export default class UniversalDiscoveryModule extends Component {
     findContentType(identifier) {
         let contentType = null;
 
-        Object.values(this.props.contentTypes).forEach((group) => {
+        Object.values(this.props.contentTypes).forEach((group: any) => {
```
tischsoic, line 229: At the beginning some type errors can be fixed by assigning `any` type, which then can be converted into something more specific in the future. As we will be adding more and more types.

src/modules/universal-discovery/universal.discovery.module.tsx
```diff
@@ -106,6 +95,20 @@ export default class UniversalDiscoveryModule extends Component {
         };
     }
 
+    private loadingBookmarksLocationsIds = {};
```
tischsoic, line 98: TS requires moving class properties from constructor to the body of a class.

src/modules/universal-discovery/universal.discovery.module.tsx
```diff
@@ -106,6 +95,20 @@ export default class UniversalDiscoveryModule extends Component {
...
+
+    private tabsLabels = {
+        browse: Translator.trans(/*@Desc("Browse")*/ 'browse.tab.label', {}, 'universal_discovery_widget'),
+        search: Translator.trans(/*@Desc("Search")*/ 'search.tab.label', {}, 'universal_discovery_widget'),
+        create: Translator.trans(/*@Desc("Create")*/ 'create.tab.label', {}, 'universal_discovery_widget'),
+        bookmarks: Translator.trans(/*@Desc("Bookmarks")*/ 'bookmarks.tab.label', {}, 'universal_discovery_widget'),
+    };
+
+    static propTypes;
+    static defaultProps;
```
tischsoic, line 110: We have to pre-declare static types.

src/modules/common/helpers/deep.clone.helper.ts
```diff
@@ -16,7 +16,7 @@ const deepClone = (data) => {
         return data;
     }
 
-    if (Object.prototype.toString.apply(data) === '[object Array]') {
+    if (isArray(data)) {
```
tischsoic, line 19: Here we needed to add custom type guard (https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types)
so that we have type checking for array inside `if` statement.
We do not need custom guards in case of simpler conditions such as `data instanceof Array`.

typings/index.d.ts
```diff
@@ -0,0 +1,13 @@
+declare global {
+    const Translator: {
```
tischsoic, line 2: Here we have type definitions of global variables so we have type checking for them as well. Without these type definitions TS shows errors. 
We can declare more types (as for `Translator`) or less (as for `eZ` below).

typings/rest-api/location.d.ts
```diff
@@ -0,0 +1,17 @@
+interface ILocation extends IRef {
```
tischsoic, line 1: 1) We can create interfaces for objects returned by REST based on XSD files: https://github.com/ezsystems/ezpublish-kernel/tree/master/doc/specifications/rest/xsd
and then use them in our application to get more type checking. 
The problem is that e.g. from REST we get location with Content containing `'_media-type': string; _href: string;` and then in JS we sometimes do: `location.Content = someFetchedContentObj`, so `location.Content` is no longer of type _IRef_. This can be solved by defining Content type in _Location_ interface as: `IRef | IContent`.

2) We should probably create namespace `eZ` for these interfaces.

3) We can also use these interfaces in jsDoc - as in `request.helper.js`.

src/modules/common/helpers/request.helper.js
```diff
@@ -2,7 +2,7 @@
  * Returns basic RequestInit object for Request
  *
  * @function getBasicRequestInit
- * @param {Object} restInfo REST config hash containing: token and siteaccess properties
+ * @param {IRestInfo} restInfo REST config hash containing: token and siteaccess properties
```
tischsoic, line 5: TS types can be used inside jsDoc, but I'm not sure if TS compiler checks them. (but editor at least seems to be checking them, so TS compiler maybe too 🤔 )
Also, what's worth to be mention is that jsDoc types no longer work if we convert file .js -> .ts.

tischsoic:
I converted some files to TS and added a few examples of types in `./typings`. In comments to this PR I tried to show required changes. Please let me know what do you think and if we want to proceed with introducing TypeScript to the project anytime soon. ping @sunpietro @dew326

PR #128, opened 2019-01-09
EZP-29777: Fixes sorting table of subitems

pawbuj:
jira : https://jira.ez.no/browse/EZP-29777


src/modules/sub-items/components/sub-items-list/sub.items.list.component.js
```diff
@@ -6,15 +6,35 @@ import GridViewComponent from '../grid-view/grid.view.component.js';
...
+}
+
 const views = {
-    table: TableViewComponent,
-    grid: GridViewComponent,
+    table: {
+        component: TableViewComponent,
+        propsTransformer: (props) => {
+
+            props.sortKey = translationsOfSortNames[Object.keys(props.sortClauses)[0]] || 'name';
```
dew326: The whole logic of sorting is inside the `table.view.component.js` maybe it would be better to have this there in the constructor? We would save calculation on each rerender and will not have dummy callback in `grid`.

tischsoic: I don't like this solution either. I agree with @dew326 that simpler solution would be to just pass `sortClauses` from _SubItemsModule_ to _ TableViewComponent_ and set defaults in the constructor:
https://github.com/ezsystems/ezplatform-admin-ui-modules/blob/75b53aa807fa18d5a94dfce19227a1c34d426e5f/src/modules/sub-items/components/table-view/table.view.component.js#L24-L25

tischsoic (changes requested):

pawbuj:
The previous version was not working correctly, I made a force push

pawbuj:
@tischsoic @dew326 The TableViewComponent has one responsibility which is to display subitems in a table. We should not add to it a second responsibility (converting sortClauses format to internal format ), because it would break the single responsibility principle. The SubItemsListComponent is responsible for displaying the proper view with proper params, so it is the right place to do props conversion.

sunpietro:
Please, generate module files

pawbuj:
https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/128#issuecomment-453063984
done

sunpietro (changes requested):
Likely, the solution is not correct. It works properly on 2.4.x

lserwatka:
Hmm, target branch is 1.3, what indicates 2.3.x, perhaps this was affected? Is this customer request?

pawbuj:
The ticket in JIRA was closed with the status "won't fix" so I am also closing this pull request.

PR #130, opened 2019-01-13
EZP-29996: Move CSS files for Universal Discovery Module to Admin UI

sunpietro:
https://jira.ez.no/browse/EZP-29996

## THINGS DONE
- [x] moved CSS styles from JS files to SASS files,
- [x] converted icons to use the `Icon` component instead of raw HTML tags,
- [x] corrected colours and sizes,
- [x] corrected clickable elements accessibility in all the UI modules.


dew326 (approved):

src/modules/universal-discovery/components/finder/finder.tree.leaf.component.js
```diff
@@ -40,11 +39,7 @@ export default class FinderTreeLeafComponent extends Component {
             return null;
         }
 
-        return (
-            <svg className="ez-icon ez-spin ez-icon-x2">
-                <use xlinkHref="/bundles/ezplatformadminui/img/ez-icons.svg#spinner" />
-            </svg>
-        );
+        return <Icon name="spinner" extraClasses="ez-spin ez-icon-x2 ez-icon--small ez-icon--light" />;
```
tischsoic, line 42: It might be good time to delete `ez-icon-x2` which is redundant and occurs in several files. But for me it's not deal-breaker. 🙂

src/modules/multi-file-upload/components/upload-list/upload.item.component.js
```diff
@@ -449,10 +450,13 @@ export default class UploadItemComponent extends Component {
...
         return (
-            <div className="c-upload-list-item__action c-upload-list-item__action--abort" onClick={this.abortUploading} title={label}>
-                <svg className="ez-icon ez-icon--small-medium">
-                    <use xlinkHref="/bundles/ezplatformadminui/img/ez-icons.svg#circle-close" />
-                </svg>
+            <div
+                className="c-upload-list-item__action c-upload-list-item__action--abort"
+                onClick={this.abortUploading}
+                title={label}
+                tabIndex="-1"
```
tischsoic, line 457: I don't think we need to set `tabIndex="-1"` for `div`s which are not focusable by default.

tischsoic (changes requested):

src/modules/multi-file-upload/components/upload-list/upload.item.component.js
tischsoic: Me and @dew326 discussed this in private and as Dariusz noted that:
> A negative value (usually tabindex="-1") means that the element should be focusable, but should not be reachable via sequential keyboard navigation. It's mostly useful to create accessible widgets with JavaScript.

source: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

tischsoic (approved):
`tabIndex="-1"` is OK, so ✅ 🙂

dew326:
@micszo changed base branch, rebased with master, fixed issue with UDW. Ready for re-test.

micszo:
Only Time TCs failed.

micszo (approved):

PR #133, opened 2019-01-18
EZP-30037: Use Webpack Encore in Admin UI modules bundle

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-30037

### DIFF
--- a/Resources/encore/ez.config.js
+++ b/Resources/encore/ez.config.js
@@ -1,0 +1,8 @@
+const path = require('path');
+
+module.exports = (Encore) => {
+    Encore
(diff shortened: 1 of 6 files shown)


tischsoic (approved):

barbaragr (changes requested):
Bookmark button in UDW is blinking:
http://g.recordit.co/FLRw0bsso0.gif

Bulk delete/move doesn't work - it gets greyed and error is displayed:
screen shot 2019-02-19 at 16 41 42

barbaragr (approved):

PR #136, opened 2019-01-29
EZP-30064: As an editor I would like to view a content tree in location view

sunpietro:
https://jira.ez.no/browse/EZP-30064

zrzut ekranu 2019-01-29 o 16 37 17

## FOR QA AND REVIEWERS:

The feature will be implemented incrementally. That means, that the design will be changing in the forthcoming pull requests. In this PR I focused on preparing the content tree module base and implementing the basic file structure.


Resources/public/scss/modules/content-tree/_list.item.scss
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    &--is-expanded {
+        > .c-list {
+            opacity: 1;
+            max-height: calculateRem(20000px);
+        }
+
+        > #{$list-item}__label {
+            #{$list-item}__toggler {
```
tischsoic, line 50: This is not compliant with BEM. We should try to avoid exceptions and use BEM whenever possible.

Resources/public/scss/modules/content-tree/_list.item.scss
```diff
@@ -0,0 +1,57 @@
...
+                    content: '';
+                }
+            }
+        }
+    }
+
+    &--is-expanded {
+        > .c-list {
+            opacity: 1;
+            max-height: calculateRem(20000px);
```
tischsoic, line 46: Do we need this? Could you explain why exactly `20000px`, please? 🙂

Resources/public/scss/modules/content-tree/_list.item.scss
sunpietro: Yes, it follows BEM convention. You need to understand what is going on here. Basically, I want to change the state of a toggler of the direct children of an item that has been toggled. I don't want to change the states of all togglers of element's children. I hope it's clear now :)

Resources/public/scss/modules/content-tree/_list.item.scss
sunpietro: I need to set a big number for height to allow height transition after expanding an element children. With this number we should be safe when a list contains a lot of items.

tischsoic: Thank you for explanation. :)

Resources/public/scss/modules/content-tree/_list.item.scss
tischsoic: I understand idea behind it, but still, I think that BEM assumes using only _class_ selectors. I thought that we are making exceptions in _admin-ui_ because of Bootstrap.

tischsoic (approved):

Resources/public/scss/modules/content-tree/_list.item.scss
sunpietro: In this case, it's also a correct BEM approach.

sunpietro:
ping @dew326

dew326 (approved):

PR #137, opened 2019-01-31
EZP-30069: As an editor I would like to change manually the width of content tree panel

sunpietro:
https://jira.ez.no/browse/EZP-30069

resize

**Requires:** https://github.com/ezsystems/ezplatform-admin-ui/pull/820


dew326 (approved):

src/modules/content-tree/content.tree.module.js
```diff
@@ -2,20 +2,79 @@ import React, { Component } from 'react';
...
+        }));
+    }
+
+    addWidthChangeListener({ nativeEvent }) {
+        const resizeStartPositionX = nativeEvent.clientX;
+        const containerWidth = parseInt(window.getComputedStyle(this._refTreeContainer.current).width, 10);
+
+        window.document.addEventListener('mousemove', this.changeContainerWidth, false);
+        window.document.addEventListener('mouseup', this.removeWidthChangeListener, false);
+        window.document.body.classList.add(CLASS_IS_TREE_RESIZING);
```
tischsoic, line 44: I would move this to another function similar to `clearDocumentResizingState` -> `addDocumentResizingState` etc.

Moreover, I think that it might be a good idea to set this listeners after setting state, like:
```js
this.setState(
    () => ({ (...), isResizing: true }), 
    () => addDocumentResizingState()
);
```

src/modules/content-tree/content.tree.module.js
```diff
@@ -2,20 +2,79 @@ import React, { Component } from 'react';
...
+        const currentPositionX = clientX;
+
+        this.setState((state) => ({
+            resizedContainerWidth: state.containerWidth + (currentPositionX - state.resizeStartPositionX),
+        }));
+    }
+
+    addWidthChangeListener({ nativeEvent }) {
+        const resizeStartPositionX = nativeEvent.clientX;
+        const containerWidth = parseInt(window.getComputedStyle(this._refTreeContainer.current).width, 10);
```
tischsoic, line 40: Isn't this the same width as in the _state_? Can we do: `const { containerWidth } = this.state`?

src/modules/content-tree/content.tree.module.js
```diff
@@ -2,20 +2,79 @@ import React, { Component } from 'react';
...
+        this.addWidthChangeListener = this.addWidthChangeListener.bind(this);
+        this.removeWidthChangeListener = this.removeWidthChangeListener.bind(this);
+        this._refTreeContainer = React.createRef();
+
         this.state = {
             selectedLocationId: props.selectedLocationId,
             locations: props.preloadedLocations,
+            resizeStartPositionX: 0,
+            containerWidth: 0,
+            resizedContainerWidth: 0,
```
tischsoic, line 21: In my opinion we don't need `resizeStartPositionX` and `resizedContainerWidth`. 
During resizing we can store container width also in `containerWidth`, which would eliminate this condition: `const width = isResizing ? resizedContainerWidth : containerWidth;` from `render` method. It would also simplify code in other places.

Also, we can replace this calculation:
```js
changeContainerWidth({ clientX }) {
        const currentPositionX = clientX;

         this.setState((state) => ({
            resizedContainerWidth: state.containerWidth + (currentPositionX - state.resizeStartPositionX),
        }));
    }
```
with:
```js
changeContainerWidth({ clientX: currentPositionX }) {
        const treeContainer = this._refTreeContainer.current;
        const containerWidth = currentPositionX - treeContainer.offsetLeft;

         this.setState(() => ({ containerWidth }));
    }
```
This would eliminate `resizeStartPositionX` and simplify code here and in other places.

But maybe `treeContainer.offsetLeft` wont work here because of some CSS styling?

src/modules/content-tree/content.tree.module.js
```diff
@@ -2,20 +2,79 @@ import React, { Component } from 'react';
...
+            isResizing: false,
         };
     }
 
+    componentWillUnmount() {
+        this.clearDocumentResizingState();
```
tischsoic, line 27: I think that we should check if we are in resizing state, otherwise there is no need to clear it.

```js
if (this.state.isResizing) {
     this.clearDocumentResizingState();
}
```

src/modules/content-tree/content.tree.module.js
```diff
@@ -2,20 +2,79 @@ import React, { Component } from 'react';
...
         super(props);
 
+        this.changeContainerWidth = this.changeContainerWidth.bind(this);
+        this.addWidthChangeListener = this.addWidthChangeListener.bind(this);
+        this.removeWidthChangeListener = this.removeWidthChangeListener.bind(this);
+        this._refTreeContainer = React.createRef();
+
         this.state = {
             selectedLocationId: props.selectedLocationId,
             locations: props.preloadedLocations,
```
tischsoic, line 18: I don't like idea of mixing logic responsible for resizing with selecting location etc. In my opinion the way to go is extracting logic responsible for resizing to another component.

tischsoic (changes requested):

src/modules/content-tree/content.tree.module.js
sunpietro: I don't see any value in moving this piece of code into a separate method. Moreover, adding event listeners and adding a CSS class to a callback invoked after updating a state doesn't change anything. It's because the listeners are added to `window.document` and `window.document.body` which are not controlled nor modified by React updates.

src/modules/content-tree/content.tree.module.js
sunpietro: No it's not the same when running this function for the first time. I don't want to get the width of a container when a component is mounted because it might be still not visible. By this approach you get a container width value when it's actually visible.

src/modules/content-tree/content.tree.module.js
sunpietro: In this piece of code I wanted to resize the container with according to a mouse pointer position. This gives more accurate results.
By storing `resizeStartPositionX` in a state while attaching `mousemove` event listeners to `window.document` allows to update the width of a container according to its starting position.

src/modules/content-tree/content.tree.module.js
sunpietro: It doesn't have much impact, because we just want to make sure the listeners and a CSS class is removed while unmounting a module. If there's no listener set then browser won't do any additional action despite of invoking this function once.

src/modules/content-tree/content.tree.module.js
sunpietro: In a context of the content tree I disagree with your approach. I would create another component only when it will be reused in other places. Currently, there's no need to do it.
There's no need to create another class/component just for it. It simplifies the way the code works without extending a bundle size (in bytes) and all the logic for the content tree module is one place.

tischsoic: We should try to adhere to single responsibility principle, this way code is clearer and easier to maintain. Clear and maintainable code is more important that few additional bytes added to bundle size.

src/modules/content-tree/content.tree.module.js
tischsoic: I understand how current algorithm works. I proposed another approach, which should work the same as current approach when it comes to final effect, but should be simpler to implement. You think that proposed approach wont work or would be more complicated?

src/modules/content-tree/content.tree.module.js
sunpietro: Until the whole functionality is big enough to extract to another component (see drag and drop), I see no point in doing it. You have to keep in mind that we're building views, not MVC and here you're going to find all the things related to that specific view.

src/modules/content-tree/content.tree.module.js
sunpietro: There's one thing you did wrong in your code sample. You assumed the currentPositionX equals to a container width.

tischsoic: No, I assumed that currentPositionX  _minus treeContainer.offsetLeft_ equals to a container width:
`containerWidth = currentPositionX - treeContainer.offsetLeft`

dew326: You are forgetting about the 3px tolerance in the resize-handler. Your approach will cause little jumps on start.

tischsoic: I thought that it is within container, but if not I think we can include it into formula, something like this, plus or minus:
`containerWidth = currentPositionX -+ resizeBarWidth - treeContainer.offsetLeft`

sunpietro: And now it's not much different from my approach.

tischsoic: It eliminates 2 state variables. I don't think that it is totally different approach, I just think that it is simpler approach.

sunpietro: Still, in this case changing it would not introduce any big value.

webhdx: spidermen

tischsoic (approved):

PR #138, opened 2019-02-06
EZP-30071: Implement responsive table in Sub-items

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-30071

## Screenshots

screen shot 2019-02-19 at 16 42 40

xj27hnmhbb


tischsoic:
This is the working proposition of Sub-items table which adjusts or can be adjusted to screen width. This is based on what I found on the Internet about adjusting tables to screens with a smaller resolution and what @inakijv sent me. Screenshots show how it looks.

Note:
Two columns - column with checkboxes and column with edit buttons - are static and do not scroll. Because of some CSS limitations, to achieve this without JS, all cells have only one line of text and constant width.
Two vertical borders separate 2 static columns from rest, but they appear also when there is no scroll - it should be possible to make them adapt with JS.

Do we want something like that or maybe just a simple scroll for the whole table, without static columns with checkboxes and buttons? ping @inakijv @SylvainGuittard @supriyabhargava ?

sunpietro:
In general, the experience is great! I really like it. But it's easy to implement when you can control the table with our own code. In case of tables generated by the backend this solution won't work. There will be an inconsistency in the app UX. We have to avoid such situations. 
cc @inakijv @SylvainGuittard

SylvainGuittard:
@sunpietro IMHO, it's fine. Sub-items list is not a regular table and is already different from the other tables in the application.

inakijv:
Great work @tischsoic! Congrats. I have added some design suggestions to your proposal.

lserwatka:
@tischsoic rebase is needed. @sunpietro @dew326 we should start slowly with CR.

tischsoic:
I finished implementation, updated PR description & screenshots, rebased branch.
I tested solution on Chrome, Firefox, Safari & Edge.

Resources/public/scss/modules/sub-items-list/_table.view.item.scss
```diff
@@ -1,5 +1,5 @@
 .c-table-view-item {
-    & + & {
+    td {
```
sunpietro: You should style it using a class selector instead of tag selector

Resources/public/scss/modules/sub-items-list/_table.view.item.scss
```diff
@@ -31,18 +33,25 @@
...
 
         &--checkbox {
-            width: 5%;
+            position: absolute;
+            left: 0;
+            width: calculateRem(30px);
+            border-top: calculateRem(1px) solid $ez-ground-primary;
+            line-height: 0;
+            padding-top: calculateRem(16px);
+            margin-top: -0.5px;     // Fixes borders alignment on Safari and Edge
```
sunpietro, line 39: `calculateRem` maybe?

Resources/public/scss/modules/sub-items-list/_table.view.scss
```diff
@@ -2,14 +2,81 @@
...
+
+    &__columns-toggler-btn {
+        padding: calculateRem(5px);
+        height: calculateRem(32px);
+        width: calculateRem(32px);
+    }
+
+    &__columns-toggler-panel {
+        border-radius: calculateRem(4px);
+        background: white;
```
sunpietro: `$ez-white`?

Resources/public/scss/modules/sub-items-list/_table.view.scss
```diff
@@ -2,14 +2,81 @@
...
+        position: absolute;
+        z-index: 10;
+        padding: calculateRem(4px) calculateRem(14px);
+        top: calculateRem(36px);
+        right: calculateRem(17px);
+        font-weight: normal;
+        box-shadow: 0 calculateRem(2px) calculateRem(4px) 0 rgba(0, 0, 0, 0.45);
```
sunpietro: `rgba($ez-black, 0.45)`

Resources/public/scss/modules/sub-items-list/_table.view.scss
```diff
@@ -36,32 +111,28 @@
...
 
         &--sort-by-name.c-table-view__head--sort-asc .c-table-view__cell--name .c-table-view__label:after,
         &--sort-by-date.c-table-view__head--sort-asc .c-table-view__cell--date .c-table-view__label:after,
         &--sort-by-priority.c-table-view__head--sort-asc .c-table-view__cell--priority .c-table-view__label:after {
-            border-left: 3px solid transparent;
-            border-right: 3px solid transparent;
-            border-bottom: 5px solid #333;
+            border-left: calculateRem(3px) solid transparent;
+            border-right: calculateRem(3px) solid transparent;
+            border-bottom: calculateRem(5px) solid #333;
```
sunpietro: `$ez-black`

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -21,6 +29,66 @@ export default class TableViewComponent extends Component {
         this.sortByPriority = this.sortByPriority.bind(this);
         this.renderItem = this.renderItem.bind(this);
         this.selectAll = this.selectAll.bind(this);
+        this.toggleColumnsToggler = this.toggleColumnsToggler.bind(this);
+        this.handleColumnsTogglerClose = this.handleColumnsTogglerClose.bind(this);
+
+        this._refColumnsTogglerButton = createRef();
+
+        this.headerLabels = {
```
sunpietro: It can be extracted outside of a class definition

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -21,6 +29,66 @@ export default class TableViewComponent extends Component {
...
+        };
+
+        this.state = {
+            isColumnsTogglerOpen: false,
+            columnsVisibility: this.getColumnsVisibilityFromLocalStorage(),
+        };
+    }
+
+    componentDidMount() {
+        document.addEventListener('click', this.handleColumnsTogglerClose, false);
```
sunpietro: `hideColumnsToggler`?

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -21,6 +29,66 @@ export default class TableViewComponent extends Component {
...
+
+        if (!isClickInsideColumnsToggler && !isClickOnColumnsTogglerBtn) {
+            this.setState(() => ({
+                isColumnsTogglerOpen: false,
+            }));
+        }
+    }
+
+    getColumnsVisibilityFromLocalStorage() {
+        try {
```
sunpietro: I'm not sure you need this `try ... catch` here, because you have control over everything here.

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -21,6 +29,66 @@ export default class TableViewComponent extends Component {
...
+            const columnsVisibilityData = localStorage.getItem(COLUMNS_VISIBILITY_LOCAL_STORAGE_DATA_KEY);
+
+            return columnsVisibilityData ? JSON.parse(columnsVisibilityData) : DEFAULT_COLUMNS_VISIBILITY;
+        } catch (error) {
+            return DEFAULT_COLUMNS_VISIBILITY;
+        }
+    }
+
+    setColumnsVisibilityInLocalStorage() {
+        try {
```
sunpietro: Same as above.

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -62,6 +130,26 @@ export default class TableViewComponent extends Component {
...
 
+    toggleColumnVisibility(column) {
+        this.setState(
+            (state) => ({
+                columnsVisibility: {
+                    ...state.columnsVisibility,
+                    [column]: !state.columnsVisibility[column],
+                },
+            }),
+            () => {
```
sunpietro: Instead of:
```
() => {
    this.setColumnsVisibilityInLocalStorage();
}
```
please do:
`this.setColumnsVisibilityInLocalStorage`

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -93,10 +182,42 @@ export default class TableViewComponent extends Component {
...
+        if (!this.state.isColumnsTogglerOpen) {
+            return null;
+        }
+
+        const { columnsVisibility } = this.state;
+
+        return (
+            <div className="c-table-view__columns-toggler-panel">
+                <ul className="c-table-view__columns-toggler-list">
+                    {Object.keys(columnsVisibility).map((columnKey) => {
```
sunpietro: Maybe `Object.entries` would work better here?

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -93,10 +182,42 @@ export default class TableViewComponent extends Component {
...
+                        const isColumnVisible = columnsVisibility[columnKey];
+                        const label = this.headerLabels[columnKey];
+
+                        return (
+                            <li
```
sunpietro: MAybe you should create a separate component so you could avoid re-renders caused by binding params to function in `onClick`. What do you think?

src/modules/sub-items/components/table-view/table.view.component.js
```diff
@@ -124,36 +245,55 @@ export default class TableViewComponent extends Component {
...
-                    </td>
-                    <td className={cellHeadClass} colSpan={2}>
-                        <span className="c-table-view__label">{headerTranslationsLabel}</span>
-                    </td>
+                    </th>
+                    <th className={cellHeadClass} />
+                    <th className={`${cellHeadClass} ${cellClass}--name ${cellSortClass}`} onClick={this.sortByName} tabIndex="-1">
+                        <span className="c-table-view__label">{this.headerLabels.name}</span>
+                    </th>
+                    {columnsVisibility.modified && (
```
sunpietro: Please, create sub-rendering methods for such cases. It will improve code readability.

src/modules/sub-items/components/table-view/table.view.item.component.js
```diff
@@ -195,30 +195,35 @@ export default class TableViewItemComponent extends PureComponent {
...
+                            {formatShortDateWithTimezone(new Date(content.lastModificationDate))}
+                        </div>
+                    </td>
+                )}
+                {columnsVisibility.contentType && (
+                    <td className="c-table-view-item__cell c-table-view-item__cell--content-type">
+                        <div className="c-table-view-item__text-wrapper">{contentTypeName}</div>
+                    </td>
+                )}
+                {columnsVisibility.priority && this.renderPriorityCell()}
```
sunpietro: This check should be inside the `this.renderPriorityCell` method

sunpietro (changes requested):

Resources/public/scss/modules/sub-items-list/_table.view.item.scss
tischsoic, line 39: This line deserves it's own comment. I had problem with alignment of borders on Edge and Safari:
screen shot 2019-02-20 at 09 17 09


### Edge
_(pictures are big, for better view open in separate tab)_
screen shot 2019-02-20 at 09 15 13

### Safari
screen shot 2019-02-20 at 09 16 25

What is more interesting, when it comes to Safari problem was occurring on Retina display but not on FullHD display. My conclusion was that it boils down to rounding of position:
screen shot 2019-02-20 at 09 16 25 copy
It seems that Safari was rounding it to 43px and `margin-top: -0.5px;` fixes this problem. This is why I also don't want this value to be in `rem`.

src/modules/sub-items/components/table-view/table.view.component.js
tischsoic: Yes, I was thinking about extracting whole toggler panel to separate component, do you like this idea?

Resources/public/scss/modules/sub-items-list/_table.view.item.scss
sunpietro: Maybe you should try adding a border on a row not on a cell itself, shouldn't you?

tischsoic: But cells with checkboxes and action buttons are outside of the row, because they have position absolute.

sunpietro: Then maybe it's not the cell that should be positioned absolutely but it's content? What do you think about creating a kind of a cell content wrapper with `div` and then position it absolutely?

tischsoic: I don't think it would solve the problem, you would still need to have border but instead on wrapper div and not on the td.

tischsoic:
PR updated according to review in *2* last commits.

lserwatka:
@sunpietro @dew326 could you make final round?

src/modules/sub-items/components/table-view/table.view.columns.toggler.list.element.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const TableViewColumnsTogglerListElement = ({ label, isColumnVisible, toggleColumnVisibility, columnKey }) => {
+    return (
+        <li className="c-table-view-columns-toggler-list-element" onClick={() => toggleColumnVisibility(columnKey)}>
```
sunpietro, line 6: I've been thinking about making it a class component so `() => toggleColumnVisibility(columnKey)` could be invoked inside a class method. That will prevent from additional re-rendering a whole component just because of creating a new function. Keep in mind that final effect of an arrow function as value of any element attribute is similar to function binding, which is a creation of a new function reference.
Sumarizing, I suggest convert this component to a class component and create a method named: `onClickHandler` or similarly:
```
onClickHandler() {
    this.props.toggleColumnVisibility(this.props.columnKey);
}
```

Does it make sense?

sunpietro (changes requested):

src/modules/sub-items/components/table-view/table.view.columns.toggler.list.element.js
tischsoic: I made it functional component because we always make them functional if they don't have state and AFAIK we quite often create functions inside functional component and then pass them to props. Another thing is, that AFAIR, a few months ago I suggested the same approach as you are suggesting now and it was rejected.

src/modules/sub-items/components/table-view/table.view.columns.toggler.js
```diff
@@ -0,0 +1,98 @@
...
+        document.removeEventListener('click', this.hidePanel);
+    }
+
+    hidePanel({ target }) {
+        if (!this.state.isOpen) {
+            return;
+        }
+
+        const isClickInsidePanel = target.closest('.c-table-view-columns-toggler__panel');
+        const isClickOnBtn = target.closest('.c-table-view-columns-toggler__btn');
```
dew326: nitpick: both are inside `c-table-view-columns-toggler` you can do closest only once.

src/modules/sub-items/components/table-view/table.view.item.component.js
```diff
@@ -172,21 +224,16 @@ export default class TableViewItemComponent extends PureComponent {
...
-        const contentTypeName = contentTypeIdentifier ? window.eZ.adminUiConfig.contentTypeNames[contentTypeIdentifier] : notAvailableLabel;
         const linkAttrs = {
             className: 'c-table-view-item__link c-table-view-item__text-wrapper',
             title: content.Name,
             href: generateLink(location.id),
         };
-        const translations = content.CurrentVersion.Version.VersionInfo.VersionTranslationInfo.Language.map((langauge) => {
-            return languages.mappings[langauge.languageCode].name;
-        });
+
```
dew326: not needed.

dew326 (approved):

katarzynazawada (changes requested):
1) Free space at the bottom of the page when only one sub-item exists
screenshot 2019-02-22 at 13 26 42

2) When clicking on Priority arrow appears also for Translations 
screenshot 2019-02-22 at 15 01 25

lserwatka:
Rebase is required.

tischsoic:
Can you please check fix from last commit, @dew326 ? It fixes 1) problem from: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/138#pullrequestreview-206832412

src/modules/sub-items/components/table-view/table.view.columns.toggler.js
```diff
@@ -28,6 +32,16 @@ export default class TableViewColumnsTogglerComponent extends Component {
         }));
     }
 
+    componentDidUpdate() {
+        const { isOpen, panelHeight } = this.state;
+
+        if (isOpen && panelHeight === null) {
```
tischsoic, line 38: Note: I'm making explicit `panelHeight === null` check in case of `this._refPanel.offsetHeight` returning `0`.

dew326 (approved):

mnocon (approved):
Looks good, issues mentioned by @katarzynazawada are gone - imho can be merged once conflicts are resolved

lserwatka:
Could you rebase?

PR #139, opened 2019-02-07
EZP-30070: As an editor I would like content tree to match design from wireframes

sunpietro:
https://jira.ez.no/browse/EZP-30070

zrzut ekranu 2019-02-7 o 12 16 07


SylvainGuittard:
@sunpietro Very good progress. Happy to see it.
Some detail: I looked a the prototype and the background color for hidden content items is only behind the text.
image

sunpietro:
@SylvainGuittard right. This is my proposal of improving consistency in the background colour rendering. As you can see, a selected content item and hovered content item get a full-width background. I followed the same rule in terms of an inactive/hidden content location.

dew326 (approved):

SylvainGuittard:
@sunpietro right! Final decision is for @inakijv ;)

PR #140, opened 2019-02-08
EZP-30066: As an editor I would like to see current location marked as selected in the content tree

sunpietro:
https://jira.ez.no/browse/EZP-30066

**Requires:** https://github.com/ezsystems/ezplatform-admin-ui/pull/836


src/modules/content-tree/components/list-item/list.item.component.js
```diff
@@ -71,9 +71,12 @@ class ListItem extends Component {
 
         return (
             <li {...itemAttrs}>
-                <a className="c-list-item__label">
-                    <span {...togglerAttrs} /> {this.renderIcon()} {name}
-                </a>
+                <div className="c-list-item__label">
+                    <span {...togglerAttrs} />{' '}
```
tischsoic: Maybe it would be better to remove `{' '}` and add some additional CSS styling instead?

src/modules/content-tree/content.tree.module.js
```diff
@@ -14,7 +14,7 @@ export default class ContentTreeModule extends Component {
         this._refTreeContainer = React.createRef();
 
         this.state = {
-            selectedLocationId: props.selectedLocationId,
+            currentLocationId: props.currentLocationId,
```
tischsoic: We need to store it in _state_ because in future it will be changing during lifetime of component?

src/modules/content-tree/components/list-item/list.item.component.js
sunpietro: That is strange, because I don't remember I added it. I'll remove it.

dew326 (approved):

tischsoic (approved):

PR #145, opened 2019-02-27
Comment out request to backend from Contnet Tree module

tischsoic:
We need to wait for backend API to be implemented in ezplatform-admin-ui repository.


src/modules/content-tree/services/content.tree.service.js
```diff
@@ -14,17 +14,17 @@ export const loadLocationItems = (parentLocationId, callback, limit = 50, offset
...
+    //         location.children = location.children.map(mapChildrenToSubitems)
 
-            return mapChildrenToSubitems(location);
-        })
-        .then(callback)
-        .catch(showErrorNotification);
+    //         return mapChildrenToSubitems(location);
+    //     })
+    //     .then(callback)
+    //     .catch(showErrorNotification);
```
dew326, line 27: Why is this catch not working? 
The result of this fetch when there is no endpoint should be a notification at the bottom.

Committing commented out code is never a good solution.

dew326 (changes requested):

PR #148, opened 2019-02-28
[WIP] EZP-30091: As an editor I would like to see which content is hidden in a given location of content tree

tischsoic:
Requires admin-ui: https://github.com/ezsystems/ezplatform-admin-ui/pull/862

## Screenshot:
txatvbedtw


tischsoic:
Feature of updating Content Tree based on visibility change in Content view won't be implemented. Page will have to be refresh to update Content Tree data.

PR #150, opened 2019-03-01
EZP-30193: As an editor I would like content tree to "Collapse all" button

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-30193

## Screenshot:
lgtuop03yv


Resources/public/scss/modules/content-tree/_list.item.scss
```diff
@@ -101,6 +101,8 @@ $list-item-height: calculateRem(20px);
         opacity: 0;
         list-style: none;
         max-height: 0;
+        max-width: 0;
```
tischsoic, line 104: Note: this fixes a bug - previously list items preserved it's width when collapsed.
Because fix was simple and short I added it in this PR.

tischsoic:
I have one doubt about how collapse should behave. Let's say we have scenario:
1) user expands item A (first 10 items are loaded)
2) user loads more items in A (A now has 20 items)
3) user clicks "Collapse all" button (all items are toggled)
4) user again expands A

What should be the result of 4th action?
1) we reload first 10 items
2) we show 10 first items but without reload
3) we show all (20) previously loaded items 

ping @SylvainGuittard @lserwatka

tischsoic:
I also see one problem with "Collapse all". At least in Chrome browser it can be covered by url popup:
screen shot 2019-03-01 at 11 36 29
It's not a big deal, but it frustrates me a little bit. 🙂 
ping @inakijv @SylvainGuittard

SylvainGuittard:
> I also see one problem with "Collapse all". At least in Chrome browser it can be covered by url popup:
> screen shot 2019-03-01 at 11 36 29
> It's not a big deal, but it frustrates me a little bit. 🙂

I see your point. The URL popup is displayed because you are on a link somewhere else on the page. If you move the cursor to `Collapse all` then the popup disappears. 
I don't know if we have to move it... Let's see what @inakijv says.

SylvainGuittard:
> I have one doubt about how collapse should behave. Let's say we have scenario:
> 
>     1. user expands item A (first 10 items are loaded)
> 
>     2. user loads more items in A (A now has 20 items)
> 
>     3. user clicks "Collapse all" button (all items are toggled)
> 
>     4. user again expands A
> 
> 
> What should be the result of 4th action?
> 
>     1. we reload first 10 items
> 
>     2. we show 10 first items but without reload
> 
>     3. we show all (20) previously loaded items
> 

Save clicks always save clicks ;)
My answer is number 3️⃣

tischsoic:
> My answer is number 3️⃣

I'm glad to hear that because this is what I've implemented, @SylvainGuittard. 😄

dew326:
So my question is I loaded more for like 200 contents and I do not need this 200 contents anymore. I want to have the default 10 again. How do I achieve that?

webhdx:
Hold on guys... Why do you think we have `load more` in the first place? This is due to possible performance issues... Now imagine situation where someone loaded 200 items by clicking `load more` many times. With approach 3️⃣ you will make a request to load 200 items in a single request...

tischsoic:
It will be reimplemented as described in 1st option. 🙂

SylvainGuittard:
Thanks @dew326 and @webhdx. You are totally right.
The answer is 1️⃣ 

Next time, I will ask the audience :)
image

lserwatka:
@tischsoic rebase is needed here

tischsoic:
I changed behaviour so that the state is reset and reloaded. Ready for review - ping @dew326

dew326 (approved):

PR #152, opened 2019-03-01
EZP-30195: As a developer I would like content tree to use config from the backend

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-30195

## Screenshot
maicft8t2l


tischsoic:
I opted for not hiding "toggle arrow" and "Show more" and I show notification when used after limit was reached. I thought that it will be less confusing for user, but maybe we should hide these UI elements and show notification just after parent was expanded in tree? ping @SylvainGuittard

Resources/translations/content_tree.en.xliff
```diff
@@ -16,6 +21,11 @@
         <target state="new">Show more</target>
         <note>key: show_more</note>
       </trans-unit>
+      <trans-unit id="3806cf634c9d1ce315ca6d9e46619b58f848ee15" resname="show_more.limit.message">
+        <source>Cannot show more. You reached the limit of loaded subitems.</source>
+        <target state="new">Cannot show more. You reached the limit of loaded subitems.</target>
```
tischsoic: Do we want to change these messages? ping @SylvainGuittard @DominikaK

webhdx:
For me it's more confusing atm :D I'd say it's better to show notification once i.e. after expanding parent and loading more items that another consecutive operation is not possible.

tischsoic:
I was thinking about it. @webhdx is right that currently it is confusing, but I believe that hiding arrow and "Show more" would be even more confusing, because both arrow and button signal to the user that there is something more beneath and in my opinion this signal should stay. 
Moreover, let's say
1) you loaded max number of subitems for location A, you get notification, OK, 
2) then you loaded max number of subitems for location B, you get notification, OK,
3) but then you refreshed page -> this results in restoring previous subtree. After this, should user get 2 notifications for A & B, or 1 for both?

But maybe I'm thinking about too extreme cases. 🤔 

Personally I would go with different design for arrow/"Show more" when limits are reached.

lserwatka:
> Personally I would go with different design for arrow/"Show more" when limits are reached.

Actually I agree with this.

inakijv:
Hi @tischsoic, 
Regarding notifications, we agree on providing feedback to the user is key here and we'd like suggest the following options according to configuration defined:

[1] `ezsettings.default.content_tree_module.load_more_limit`
Default number of items loaded every time. No need of a message.

[2] `ezsettings.default.content_tree_module.children_load_max_limit`
User will see a message displayed instead of load more button (see wireframe below with _loading limit reached_ message displayed).

15_content_tree


[3] `ezsettings.default.content_tree_module.tree_max_depth`
User will see a notification bar message when content tree has reached its total loading capacity just right after clicking on elements that trigger actions that would go over content tree defined configuration (see wireframe below with Notification bar message displayed right after user clicked on).

16_content_tree


[4] `ezsettings.default.content_tree_module.allowed_content_types`
& `ezsettings.default.content_tree_module.ignored_content_types`
No need of a design/notification because items will or will not be displayed.

DominikaK:
Not sure about the `Loading limit reached` message, it sounds like a limit for the whole tree, not just a specific parent. Maybe something like `Limit for this parent reached`?

DominikaK (approved):
+1 for strings

Resources/translations/content_tree.en.xliff
```diff
@@ -11,6 +11,11 @@
         <target state="new">Collapse all</target>
         <note>key: collapse_all</note>
       </trans-unit>
+      <trans-unit id="0cba8ae8f7a758a11886b6be4e7ce7e570063f8f" resname="expand_item.limit.message">
+        <source>Cannot load subitems for given location due to performance limits. You reached max tree depth.</source>
```
tischsoic, line 15: We changed this message with @DominikaK so that it is more precise. Thank you Dominika for helping me. :) 
ping @inakijv @SylvainGuittard

tischsoic:
Behaviour changed according to: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/152#issuecomment-469088321
Ready for review - ping @dew326

Resources/public/scss/modules/content-tree/_list.item.scss
```diff
@@ -91,6 +90,23 @@ $list-item-height: calculateRem(20px);
         .ez-icon {
             fill: $ez-color-base-dark;
         }
+
+        &--loading-limit-reached {
+            color: $ez-secondary-ground;
+            font-style: italic;
+
+            &.btn[disabled] {
+                opacity: 1;
```
dew326: Why we want the disabled button to have full opacity? The lower opacity was added to show that you cannot click on button.

Resources/public/scss/modules/content-tree/_list.item.scss
```diff
@@ -77,7 +77,6 @@ $list-item-height: calculateRem(20px);
         border-radius: calculateRem(4px);
         margin-left: calculateRem(26px);
         transition: background 0.3s $ez-admin-transition, color 0.3s $ez-admin-transition;
-        cursor: pointer;
```
dew326, line 80: Why not pointer anymore?

tischsoic: I realised that, it is added by default to buttons and we do not need this to have this style.

Resources/public/scss/modules/content-tree/_list.item.scss
tischsoic: Because in wireframes we have specific color and with opacity 0.3 color is different.

On the second thought, maybe better approach would be to show "Load limit reached" information as separate div, not "Show more" button. We would avoid this weird styling + the logic in `renderLoadMoreBtn` would be simpler. What do you think?

dew326: Seems like a good idea.

tischsoic:
Fixed according to Code Review - ping @dew326

dew326 (approved):

Resources/translations/content_tree.en.xliff
SylvainGuittard: Yesterday, I suggested to Iñaki:
`For performance reasons, the navigation using content tree is limited. Please use the sub-items section to continue navigation`
What do you think?

PR #155, opened 2019-03-04
EZP-30205: As an editor I would like content tree to hold separate state for different users

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-30205

**Requires:** https://github.com/ezsystems/ezplatform-admin-ui/pull/871


lserwatka:
rebase is needed

PR #156, opened 2019-03-06
EZP-30232: Item highlight is cropped in Content Tree with scroll

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-30232

### Before
screen shot 2019-03-06 at 09 58 48

### After
screen shot 2019-03-06 at 12 48 52

### DIFF
--- a/Resources/public/scss/modules/content-tree/_list.scss
+++ b/Resources/public/scss/modules/content-tree/_list.scss
@@ -3,3 +3,10 @@
     padding: 0 0 0 calculateRem(12px);
     list-style: none;
 }
+
+.m-tree__scrollable-wrapper {
+    > .c-list {
+        display: table;
+        width: 100%;
+    }
+}


Resources/public/scss/modules/content-tree/_list.scss
```diff
@@ -3,3 +3,10 @@
     padding: 0 0 0 calculateRem(12px);
     list-style: none;
 }
+
+.m-tree__scrollable-wrapper {
+    > .c-list {
+        display: table;
```
tischsoic, line 9: I also tried solutions with 
- `display: inline-block`
- `display: flex; align-items: stretch;` on `.m-tree__scrollable-wrapper`
but they didn't work. I know that _table layouts_ are taboos in CSS world but this solutions seems to work perfectly.

If you have better idea @dew326 I will gladly hear it. 🙂

tischsoic: Do we want to move this to `_main.scss` or leave it here? I added it here because it styles _c-list_.

dew326: I think adding hover state on `c-list-item__link` should solve the problem.

tischsoic: It's interesting idea, but it won't work with shorter items - we want all highlights to stretch to the right side of the container. Problematic example:

screen shot 2019-03-06 at 14 21 32

screen shot 2019-03-06 at 14 18 25

dew326 (approved):

micszo (approved):

PR #159, opened 2019-03-12
EZP-30289: Content Tree does not load when root location has no items

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-30289


dew326 (approved):

src/modules/content-tree/content.tree.module.js
```diff
@@ -15,6 +15,7 @@ export default class ContentTreeModule extends Component {
         this.handleCollapseAllItems = this.handleCollapseAllItems.bind(this);
         this.limitSubitemsInSubtree = this.limitSubitemsInSubtree.bind(this);
 
+        try {
```
tischsoic, line 18: I decided to add try/catch in case anyone has corrupted saved subtree it will fallback to default. Are you fine with this, ping @dew326 ?

dew326: Well, we should avoid saving the corrupted data in the first place. Why the corrupted data was saved?

tischsoic: Corrupted data was saved because of error, which this PR fixes.

dew326 (approved):

m-tyrala (approved):
~~QA approve~~

Or maybe not - behat tests are failing - cant create Content Types through api, or there is some cache problem and they are not visible in content types list.

https://res.cloudinary.com/ezplatformtravis/image/upload/v1552466360/screenshots/5c88c1b893c5e315027748-vendor_ezsystems_ezplatform-admin-ui_features_contenttypefields_feature_34_qkv4rw.png

lserwatka:
@m-tyrala could you restart it? Does it pass locally?

m-tyrala (approved):
That's the problem we have in master also, so I can approve this PR.

lserwatka:
Could you report it?

m-tyrala:
Sure, I'm on it.

PR #165, opened 2019-03-21
EZP-30154: Cannot select an item when browsing the UDW if my role has a policy Content/Create with Content Type limitation

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-30154


src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -607,10 +607,8 @@ export default class UniversalDiscoveryModule extends Component {
         const isAlreadySelected = selectedContent.find((item) => item.ContentInfo.Content._id === data.ContentInfo.Content._id);
         const isOverLimit = !!this.props.selectedItemsLimit && selectedContent.length >= this.props.selectedItemsLimit;
         const contentTypeInfo = contentTypesMap[data.ContentInfo.Content.ContentType._href];
-        const isAllowedContentType =
-            !this.props.cotfAllowedContentTypes.length || this.props.cotfAllowedContentTypes.includes(contentTypeInfo.identifier);
 
-        if (isAlreadySelected || isOverLimit || !isAllowedContentType) {
```
dew326, line 613: Is this fix is about removing the feature of checking if content type can be selected? It doesn't seem like a proper fix.

dew326 (changes requested):

tischsoic:
Problem is going to be resolved in other PR: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/175

PR #167, opened 2019-03-22
EZP-30332: [Safari] Bookmark button is not visible before clicking

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-30332


dew326 (approved):

tischsoic:
This PR does not fixes Bookmark star in 100% (on the Safari), the light grey background sometimes does not appear, but this problem occurs also in 2.4, so I believe requires another PR with base 2.4.

I also checked bookmark star from location page (ezplatform-admin-ui) and it also has problems with background (on the Safari):
TjDKODvIGg

My proposition is to make follow-up for background problem. ping @lserwatka @barbaragr

barbaragr (approved):

PR #169, opened 2019-03-22
EZP-30312: Performing actions from subitems list does not refresh the Content Tree

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-30312

Related PR from admin-ui for location visibility change: https://github.com/ezsystems/ezplatform-admin-ui/pull/921


src/modules/content-tree/content.tree.module.js
```diff
@@ -43,6 +44,8 @@ export default class ContentTreeModule extends Component {
         loadSubtree(this.props.restInfo, this.subtree, (loadedSubtree) => {
             this.setInitialItemsState(loadedSubtree[0]);
         });
+
+        document.body.addEventListener('ctRefresh', this.refreshContentTree, false);
```
tischsoic: I've chosen `ctRefresh` based on names such as: `pbIframeLoaded`, or` fbFormBuilderUnloaded`, but I don't think it is perfect. What do you think @dew326 ?

tischsoic:
As agreed with @SylvainGuittard, Content Tree will refresh on bulk delete/move and location visibility change.

src/modules/content-tree/content.tree.module.js
dew326: I think we should prefix events with `ez-` like we have with `ez-notify` or `ez-bookmark-change` so for me the name should be `ez-content-tree-refresh`

src/modules/content-tree/content.tree.module.js
```diff
@@ -43,6 +44,8 @@ export default class ContentTreeModule extends Component {
         loadSubtree(this.props.restInfo, this.subtree, (loadedSubtree) => {
             this.setInitialItemsState(loadedSubtree[0]);
         });
+
+        document.body.addEventListener('ez-content-tree-refresh', this.refreshContentTree, false);
```
dew326: This `componentDidMount()` has the early return. There is a possibility this event will not be attached. Seems like should be before the return or am I missing something?

tischsoic: You're right, good catch. :)

dew326 (approved):

katarzynazawada (approved):

PR #172, opened 2019-03-27
EZP-30368: As an editor I would like Content Tree to display message when there are no sub-items

tischsoic:
JIRA: https://jira.ez.no/browse/EZP-30368

## Screenshot
Screen Shot 2019-03-28 at 09 06 38


lserwatka:
Design should be updated, this looks ugly. Ping @inakijv

Resources/public/scss/modules/content-tree/_main.scss
```diff
@@ -41,4 +41,10 @@
         color: $ez-color-base-light;
         cursor: pointer;
     }
+
+    &__no-items-message {
+        padding: calculateRem(10px);
+        font-size: calculateRem(20px);
+        color: $ez-color-base-light;
+    }
```
inakijv, line 50: - For this kind of messages we are using `color: $ez-color-base-dark;`.
- Font-size, please change it to 12px (the rule we are following is that the size of text for this message is the same as when regular content is displayed).
- Please add `font-style: italic;`.

inakijv (commented):
Thank you @tischsoic. Added just a small comment.

src/modules/content-tree/components/content-tree/content.tree.js
```diff
@@ -61,6 +61,19 @@ export default class ContentTree extends Component {
         window.document.body.classList.remove(CLASS_IS_TREE_RESIZING);
     }
 
+    renderNoSubitemMessage() {
+        const { items } = this.props;
+        const rootLocation = items[0];
+        const itemsLoaded = !!rootLocation;
```
dew326: As I see you need a negation of `itemsLoaded` so maybe just `const itemsNotLoaded = !rootLocation`?
But now I'm wondering why you want to return if items are not loaded and checking the root? Probably the name should be `rootNotLoaded`?

dew326 (changes requested):

src/modules/content-tree/components/content-tree/content.tree.js
tischsoic: Name was confusing, I changed it and deleted double `!!` because they are not necessary in this case. 
I didn't opt for using `not` negation in variable name because I find it hard to process in my mind when analysing code.

tischsoic:
I updated design based on @inakijv suggestions. ping @inakijv @lserwatka

dew326 (approved):

barbaragr (approved):

PR #179, opened 2019-04-09
EZP-29094: Cannot create user from Content on the Fly

pawbuj:
jira : https://jira.ez.no/browse/EZP-29094
related to : https://github.com/ezsystems/ezplatform-admin-ui/pull/953


src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -24,7 +24,19 @@ export default class ContentCreatorComponent extends Component {
...
     handlePublish() {
         this._refIframe.contentWindow.onbeforeunload = () => {};
-        this._refIframe.contentWindow.document.body.querySelector('#ezrepoforms_content_edit_publish').click();
+        this.submitForm();
+    }
+
+    submitForm() {
+        const contentEditSubmit = this._refIframe.contentWindow.document.body.querySelector('#ezrepoforms_content_edit_publish');
+        if(contentEditSubmit) {
+            contentEditSubmit.click();
```
dew326: I think it's not a good approach. This will submit all existing forms. 
I think we should add some selector to the submit button in both forms like `data-action="submit"` or something like that and query for this selector.

dew326 (changes requested):

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -24,7 +24,20 @@ export default class ContentCreatorComponent extends Component {
...
+
+    submitForm() {
+        const selectors = ['#ezrepoforms_content_edit_publish', '#ezrepoforms_user_create_create'];
+
+        for(let selector of selectors) {
+            let submit = this._refIframe.contentWindow.document.body.querySelector(selector);
+            if(submit) {
+                submit.click();
+
+                return;
```
dew326: Did you try this in the UI?

dew326: Nevertheless, this is not what I suggested. We should not rely on "if exists" but rather have some special selector as I proposed in the previous comment.

pawbuj: > Did you try this in the UI?

Yes I did, works like a charm

pawbuj: "if exists" is always required if we want to avoid errors in the console. I really do not see the difference, 
we have a special selector even two. This is not a script for universal purpose but rather a special one that supports the process of creating content on the fly. In this procees, we support 2 forms only. I my opinion is better to list those forms explicitly as is done currently.

dew326: For maintenance purposes. If you add one common selector you will no need to come into this JS when you want to add support for next form.

In my case, it's not like a charm :)
Screenshot at Apr 11 10-37-29

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -24,7 +24,20 @@ export default class ContentCreatorComponent extends Component {
...
     handlePublish() {
         this._refIframe.contentWindow.onbeforeunload = () => {};
-        this._refIframe.contentWindow.document.body.querySelector('#ezrepoforms_content_edit_publish').click();
+        this.submitForm();
+    }
+
+    submitForm() {
+        const selectors = ['#ezrepoforms_content_edit_publish', '#ezrepoforms_user_create_create'];
+
+        for(let selector of selectors) {
```
tischsoic: What do you think about:
```js
const form = this._refIframe.contentWindow.document.body.querySelector('#ezrepoforms_content_edit_publish, #ezrepoforms_user_create_create');

form.click();
```
? Assuming that one of these forms exists _querySelector_ will return it.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
pawbuj: Your code is a bit different

dew326: Yes, it is. What I meant it should be `break;` not `return;`, empty `return;` at the end of the method is really new for me.
But I still opt for one selector, not two.

pawbuj: Changed, new implementation is totally maintenance-free.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -24,7 +24,15 @@ export default class ContentCreatorComponent extends Component {
 
     handlePublish() {
         this._refIframe.contentWindow.onbeforeunload = () => {};
-        this._refIframe.contentWindow.document.body.querySelector('#ezrepoforms_content_edit_publish').click();
+        this.submitForm();
+    }
+
+    submitForm() {
+        const submit = this._refIframe.contentWindow.document.body.querySelector('[type="submit"]');
```
dew326: This will send the first form which has a button type submit. As I suggested we need custom data-attribute.

dew326 (changes requested):

src/modules/universal-discovery/components/content-creator/content.creator.component.js
pawbuj: yes, because by design it can be only one form with just one submit button and this is guaranteed by the backend. Adding additional special selector just worsen developer experience.

dew326: No, it's not guaranteed. This is fully extensible.

lserwatka: I agree with @dew326 This is too generic. We need to be precise which form are going to submit.

Nattfarinn: @pawbuj Form can have multiple submit buttons and often does, especially regarding content creation. This form is fully extensible and different submit values are used to generate different events (used by FormProcessors in RepositoryFormsBundle).

Example ref. https://github.com/ezsystems/repository-forms/blob/master/bundle/Controller/ContentEditController.php#L114

Here's Symfony Doc on how to extend existing form: https://symfony.com/doc/3.4/form/create_form_type_extension.html

And here you have our own extension that provides additional Submit button to content edit form: https://github.com/ezsystems/flex-workflow/blob/master/lib/Core/Form/ContentEditTypeExtension.php

pawbuj: @Nattfarinn You misunderstand me, I am not saying in general of course :) but about this particular case: "creating content via on the fly modal". We use here just 2 forms, one in the same time 
https://github.com/ezsystems/ezplatform-admin-ui/blob/554d06c0a5b3d381c2d6a2485b0bd5b29654bbfe/src/bundle/Controller/ContentOnTheFlyController.php#L107
https://github.com/ezsystems/ezplatform-admin-ui/blob/EZP-29094/src/bundle/Controller/ContentOnTheFlyController.php#L143

each of those forms has just one submit button 
https://github.com/ezsystems/ezplatform-admin-ui/blob/554d06c0a5b3d381c2d6a2485b0bd5b29654bbfe/src/bundle/Resources/views/content/content_on_the_fly/content_create_on_the_fly.html.twig#L30-L36
and this is guaranteed by the backend.

To finish the discussion, I will add an additional selector.

Nattfarinn: > We use here just 2 forms, one in the same time (...) each of those forms has just one submit button (...) and this is guaranteed by the backend.

I am not sure what you mean by "each of those forms has just one submit button (...) and this is guaranteed by the backend". Each of those forms **have** *multiple* hidden submit buttons. Even with `drafts_enabled: false` there is also FlexWorkflowBundle, DateBasedPublisherBundle, preview and workflow. Each of these features adds at least one submit button to `ContentEditType` which is used in Content on the Fly as well.

You should never rely on *get me first submit button* if there is a way to add/remove/rearrange buttons as you like. You are missing the point we provide fully extensible software while we keep a Backward Compatibility on our features. If it is **extensible** it is **not** guaranteed by the backend.

Screenshot 2019-04-12 at 11 19 28

dew326:
If this is for 2.5 you need to include compiled files.

src/modules/universal-discovery/components/content-creator/content.creator.component.js
```diff
@@ -24,7 +24,15 @@ export default class ContentCreatorComponent extends Component {
 
     handlePublish() {
         this._refIframe.contentWindow.onbeforeunload = () => {};
-        this._refIframe.contentWindow.document.body.querySelector('#ezrepoforms_content_edit_publish').click();
+        this.submitForm();
+    }
+
+    submitForm() {
+        const submit = this._refIframe.contentWindow.document.body.querySelector('[data-action="submit"]');
```
ViniTou: I think we should use more meaningful name and maybe move this to correct `Type` class?

Nattfarinn: Not sure what you mean by that @ViniTou , but I would just change the `submit` value to `publish`.

ViniTou: @Nattfarinn 
to add it here instead of twig view.
`\EzSystems\RepositoryForms\Form\Type\Content\ContentEditType`

vs

https://github.com/ezsystems/ezplatform-admin-ui/pull/953/files#diff-529cccd697dbcafb9e85e279bbf542aeR32

dew326 (approved):

micszo (approved):
Works fine. 👍 
Also retested other CTs in COTF.

PR #181, opened 2019-04-10
Delete JS build files and webpack config

tischsoic:
**They are no longer needed since we have Webpack Encore.**


sunpietro (changes requested):
We have to think about moving the Babel config to meta. Otherwise, we are not able to control the output, because the default Babel config for Webpack Encore could be totally different from our needs that will end up with bigger files in output and with less possibilities regarding the usage of ESx features in our code (because some Babel plugins might not be included by default).

dew326 (approved):

sunpietro (approved):

PR #185, opened 2019-04-30
EZP-30506 : Add children count column in sub-items tab

guillaumeogier:
We add the children count column in the sub-items tab. We add the show / hide option on this new column.

JIRA : [https://jira.ez.no/browse/EZP-30506](https://jira.ez.no/browse/EZP-30506) 

image


SylvainGuittard:
@guillaumeogier thanks for your contribution.
Could you please add a screenshot to the description?
Thanks

PR #186, opened 2019-05-10
EZP-30530: Create a base for the new UDW

sunpietro:
https://jira.ez.no/browse/EZP-30530

## Things done

- [x] updated node modules dependencies,
- [x] introduced Jest testing framework,
- [x] created `BaseTabComponent` that provides features for content selection, displaying content meta preview and bookmarking content,
- [x] re-implemented the UDW module to render tabs based on a provided configuration. There are no default tabs by now,
- [x] created the `LoadingSpinnerComponent` that can be re-used in multiple places in our codebase,
- [x] created the `classnames` helper to allow adding CSS classes based on provided conditions,
- [x] created the `rastInfo` helper to get REST meta info from HTML meta tags.

Zrzut ekranu 2019-05-10 o 08 26 57

**Visual aspects are in progress and will be finalised in the final phase of development**


src/modules/common/loading-spinner/loading.spinner.component.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../icon/icon';
+
+const LoadingSpinnerComponent = ({ extraClasses }) => {
+    const iconClasses = `c-loading-spinner ez-spin ez-icon-x2 ez-icon-spinner ${extraClasses}`;
```
dew326: Just a nitpick. You can use here the helper. Currently, this will have a trailing space in most cases, it's ugly in HTML :P

src/modules/common/classnames/classnames.js
```diff
@@ -0,0 +1,11 @@
+export const classnames = (classes) => {
```
dew326: I know the name is taken from the package but I think we should use camelCase here? If so the file name should also be changed to `class-names`.

src/modules/common/popup/popup.component.js
```diff
@@ -15,161 +18,110 @@ const MODAL_SIZE_CLASS = {
...
-        modal.one('hidden.bs.modal', this.props.onClose);
-    }
-
-    onKeyUp(event) {
-        const { originalEvent } = event;
-        const escKeyCode = 27;
-        const escKeyPressed = originalEvent && (originalEvent.which === escKeyCode || originalEvent.keyCode === escKeyCode);
+const Popup = (props) => {
+    const { footerChildren, additionalClasses, size, noHeader, children } = props;
+    const { hasFocus, onClose, title, subtitle } = props;
```
dew326, line 23: Why two `const` on `props`?

src/modules/common/popup/popup.component.js
```diff
@@ -15,161 +18,110 @@ const MODAL_SIZE_CLASS = {
...
+        if (isVisible) {
+            document.body.classList.add(CLASS_MODAL_OPEN);
+            document.body.classList.add(CLASS_NON_SCROLLABLE);
+
+            modal.on('keyup', onKeyUp);
+            modal.one('hidden.bs.modal', onClose);
+        } else {
+            modal.modal('hide');
+
+            window.document.body.classList.remove(CLASS_MODAL_OPEN, CLASS_NON_SCROLLABLE);
```
dew326, line 98: just `document`

src/modules/udw/services/universal.discovery.service.js
```diff
@@ -0,0 +1,243 @@
...
+};
+
+/**
+ * Loads location
+ *
+ * @function loadLocation
+ * @param {Object} params params hash containing REST config: token and siteaccess properties; locationId and offset
+ * @param {Function} callback
+ */
+export const loadLocation = ({ token, siteaccess, locationId, limit = QUERY_LIMIT, offset = 0 }, callback) => {
```
dew326, line 82: As far as I remember this endpoint can load multiple locations when you pass location comma separated. So probably we should name this `loadLocations`

src/modules/udw/services/universal.discovery.service.js
```diff
@@ -0,0 +1,243 @@
...
+};
+
+/**
+ * Finds locations related to the parent location
+ *
+ * @function findLocationsByParentLocationId
+ * @param {Object} params params hash containing REST config: token and siteaccess properties; parentLocationId and offset; sortClauses
+ * @param {Function} callback
+ */
+export const findLocationsByParentLocationId = (
```
dew326, line 144: The same it can find for multiple locations

src/modules/udw/services/universal.discovery.service.js
```diff
@@ -0,0 +1,243 @@
...
+/**
+ * Finds content matching a given text query
+ *
+ * @function findContentBySearchQuery
+ * @param {Object} restInfo REST config hash containing: token and siteaccess properties
+ * @param {String} query
+ * @param {Function} callback
+ * @param {Number} limit
+ */
+export const findContentBySearchQuery = ({ token, siteaccess }, query, callback, limit = QUERY_LIMIT) => {
```
dew326, line 191: probably `findLocationsBySearchQuery`

src/modules/udw/udw.module.js
```diff
@@ -0,0 +1,120 @@
...
+        return total;
+    }, {});
+
+    return contentTypesMap;
+};
+
+const UDWModule = ({ title, onClose, tabs, maxHeight }) => {
+    const [contentTypesMap, setContentTypesMap] = useState(null);
+    // useReducer maybe?
+    const [state, setState] = useState(() => {
```
dew326: Is it intentional to have here `state` instead of `activeTabId` like in the other places?

src/modules/common/popup/popup.component.js
sunpietro: I did it for better code readability. It was easier to scan narrower lines of code in these cases.

src/modules/common/popup/popup.component.js
sunpietro: I follow the convention we agreed some time ago when implementing self-invoking functions on load:
```
(function (global, doc) {})(window, window.document);
```

src/modules/udw/services/universal.discovery.service.js
sunpietro: I will correct it. It was just a copy-paste from the `src/modules/universal-discovery` directory

src/modules/udw/udw.module.js
sunpietro: Good catch!

dew326 (approved):

package.json
```diff
@@ -1,32 +1,39 @@
...
         "build": "webpack --config webpack.prod.js",
         "builddev": "webpack --config webpack.dev.js",
         "watch": "webpack --watch --config webpack.dev.js",
         "watchprod": "webpack --watch --config webpack.prod.js",
-        "fix-cs": "eslint --fix src/ && prettier --write src/**/*.js"
+        "fix-cs": "eslint --fix src/ && prettier --write src/**/*.js",
+        "test": "jest --config ./jest.config.js"
+    },
+    "dependencies": {
+        "react": "^16.2.0",
```
tischsoic, line 13: I think we should raise React version number to 16.8 as we are now using hooks so 16.8 is minimum.

src/modules/common/css-class-names/css.class.names.js
```diff
@@ -0,0 +1,11 @@
+export const createCssClassNames = (classes) => {
+    if (Object.prototype.toString.call(classes) !== '[object Object]') {
```
tischsoic, line 2: What I find handy in classnames package (https://www.npmjs.com/package/classnames) is possibility to pass string:
```js
 const className = classNames('c-month-day-view', {
        'c-month-day-view--past': isPast,
    });
```
and I would like to see similar functionality here. :) Ability to pass array would be also helpful. I'm also not convinced if we really need to avoid adding dependency on https://www.npmjs.com/package/classnames . From what I've checked size difference won't be big.

src/modules/udw/tabs/base.tab.component.js
```diff
@@ -0,0 +1,72 @@
...
+import { loadContentInfo } from '../services/universal.discovery.service';
+import { restInfo } from '../../common/rest-info/rest.info';
+import deepClone from '../../common/helpers/deep.clone.helper';
+
+const BaseTabComponent = ({ children }) => {
+    const contentTypesMap = useContext(ContentTypesContext);
+    const [showContentMetaPreview, setShowContentMetaPreviewState] = useState(false);
+    const [contentMeta, setContentMeta] = useState(null);
+    const [selectedContent, setSelectedContent] = useState([]);
+    const addContentTypeInfoToItem = (item) => {
```
tischsoic, line 13: We may consider moving some functions which does not use state, context etc. outside of react component. If done right it should improve code readability. :)
Now it is a little bit hard to read, at least here on Github. :P

sunpietro: Good point. Usually, I follow the convention of extracting non-component related code outside of component. In this case, I forgot about it. Will be fixed in one of the next PRs.

src/modules/common/css-class-names/css.class.names.js
sunpietro: I didn't like the approach coming from a mentioned library.
I prefer the way it is now because thanks to this approach we're not confusing developers with setting multiple params. There's only one param that takes an object with CSS class names as keys and conditions as values. Giving `true` as a value for a CSS class name, that has to be added always, is a really small overhead in my opinion. Compared to complexity of 2 params and 2 different approaches of setting a class name on an element.
What your take on this @dew326 ?

dew326: For me also too many arguments makes this method unnecessary complex.

tischsoic: OK, I'm glad I could hear your opinion. :)

PR #187, opened 2019-05-10
EZP-30525: Convert Browse Tab as a standalone ReactJS module

sunpietro:
https://jira.ez.no/browse/EZP-30525

## Things done

- [x] converted the Browse tab into a standalone module,
- [x] implemented the render as asson as possible approach. The missing content meta will be loaded in the meantime when displaying partially loaded content meta preview. It reduces a delay between selecting a content and displaying the content meta preview.

### DIFF
--- a/src/modules/udw/common/select-content-button/select.content.button.component.js
+++ b/src/modules/udw/common/select-content-button/select.content.button.component.js
@@ -1,0 +1,74 @@
+import React, { useState, useEffect, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../../../common/icon/icon';
+import { createCssClassNames } from '../../../common/css-class-names/css.class.names';
+
+const SelectContentButtonComponent = ({ checkCanSelectContent, location, onSelect, onDeselect, isSelected }) => {
+    const [isSelectContentEnabled, setIsSelectContentEnabled] = useState(true);
+    const handleSelect = useCallback(
+        (event) => {
+            event.stopPropagation();
+
+            onSelect(location);
+        },
+        [location, onSelect]
+    );
+    const handleDeselect = useCallback(
+        (event) => {
+            event.stopPropagation();
+
+            onDeselect(location.id);
+        },
+        [location, onDeselect]
+    );
+    const toggleEnabledState = useCallback(
+        (selectContentEnabled) => {
+            if (isSelectContentEnabled === selectContentEnabled) {
+                return;
+            }
+
+            setIsSelectContentEnabled(selectContentEnabled);
+        },
+        [isSelectContentEnabled]
+    );
+
+    useEffect(() => {
+        checkCanSelectContent(location, toggleEnabledState);
+    }, [checkCanSelectContent, location, toggleEnabledState]);
+
+    const iconId = isSelected ? 'checkmark' : 'create';
+    const attrs = {
+        type: 'button',
+        className: createCssClassNames({
+            'c-select-content-button': true,
+            'c-select-content-button--selected': isSelected,
+        }),
+        onClick: isSelected ? handleDeselect : handleSelect,
+    };
+
+    if (!isSelected && !isSelectContentEnabled) {
+        return null;
+    }
+
+    return (
+        <button {...attrs}>
+            <Icon name={iconId} extraClasses="ez-icon--small" />
+        </button>
+    );
+};
+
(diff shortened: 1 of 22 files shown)


src/modules/udw/common/select-content-button/select.content.button.component.js
```diff
@@ -0,0 +1,74 @@
...
+};
+
+SelectContentButtonComponent.propTypes = {
+    checkCanSelectContent: PropTypes.func.isRequired,
+    location: PropTypes.shape({
+        id: PropTypes.string.isRequired,
+    }).isRequired,
+    onSelect: PropTypes.func.isRequired,
+    onDeselect: PropTypes.func.isRequired,
+    isSelected: PropTypes.bool.isRequired,
```
dew326: It has default so it's not required.

src/modules/udw/common/selected-content/selected.content.item.component.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../../../common/icon/icon';
+
+const notAvailableLabel = Translator.trans(/*@Desc("N/A")*/ 'select_content.not_available.label', {}, 'universal_discovery_widget');
+const SelectedContentItemComponent = ({ contentName, locationId, contentTypeIdentifier, contentTypeName, onRemove }) => {
+    let icon = null;
+
+    if (contentTypeIdentifier) {
+        icon = <Icon name={contentTypeIdentifier} extraClasses="c-selected-content-item__icon ez-icon--small" />;
```
dew326: Why not to use the `ContentTypeIconComponent`?

src/modules/udw/services/bookmark.service.js
```diff
@@ -0,0 +1,106 @@
...
+ * @param {Function} callback
+ */
+export const checkIsBookmarked = (restInfo, locationId, callback) => {
+    const basicRequestInit = getBasicRequestInit(restInfo);
+    const request = new Request(`${ENDPOINT_BOOKMARK}/${locationId}`, {
+        ...basicRequestInit,
+        method: 'HEAD',
+    });
+    const bookmarkedStatusCode = 200;
+    const notBookmarkedStatusCode = 404;
```
dew326, line 97: Can we change this and use always 200 with the response `{ bookmarked: true|false}`?

sunpietro: Good idea, but I'll keep it like this, Because we need changes on the backend side as well. This work has not been planned yet.

sunpietro:
ping @dew326

src/modules/udw/common/select-content-button/select.content.button.component.js
```diff
@@ -0,0 +1,74 @@
...
+                return;
+            }
+
+            setIsSelectContentEnabled(selectContentEnabled);
+        },
+        [isSelectContentEnabled]
+    );
+
+    useEffect(() => {
+        checkCanSelectContent(location, toggleEnabledState);
```
tischsoic, line 36: Should we `setIsSelectContentEnabled(false)` before `checkCanSelectContent`?

src/modules/udw/common/selected-content/selected.content.component.js
```diff
@@ -0,0 +1,93 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import SelectedContentItemComponent from './selected.content.item.component';
+import PopupComponent from '../../../common/tooltip-popup/tooltip.popup.component';
+import { createCssClassNames } from '../../../common/css-class-names/css.class.names';
+
+const noConfirmedContentTitle = Translator.trans(
```
tischsoic: some variables holding translations text use UPPER_CASE, and here we have camelCase, is this intended? :)

src/modules/udw/common/select-content-button/select.content.button.component.js
sunpietro: This is not the best idea in the world, because it will cause extra re-renders of a component. Another effect is basing on changes of `isSelectContentEnabled` flag.

src/modules/udw/common/selected-content/selected.content.component.js
sunpietro: Good catch. I'll correct that here and there.

tischsoic (approved):

PR #188, opened 2019-05-10
EZP-30526: Convert Search Tab as a standalone ReactJS module

sunpietro:
https://jira.ez.no/browse/EZP-30526

Requires https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/187 to be merged first.

Zrzut ekranu 2019-05-10 o 09 45 17

### DIFF
--- a/src/modules/udw/tabs/search/components/content-table/content.table.component.js
+++ b/src/modules/udw/tabs/search/components/content-table/content.table.component.js
@@ -1,0 +1,127 @@
+import React, { useState, useCallback, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import ContentTablePaginationComponent from './content.table.pagination.component';
+import ContentTableItemComponent from './content.table.item.component';
+import ContentTableHeaderComponent from './content.table.header.component';
+import LoadingSpinnerComponent from '../../../../../common/loading-spinner/loading.spinner.component';
+
+const splitToPages = (items, perPage) => {
+    return items.reduce((pages, item, index) => {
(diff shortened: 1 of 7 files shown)


src/modules/udw/tabs/search/components/content-table/content.table.component.js
```diff
@@ -0,0 +1,129 @@
...
+    const renderPage = () => {
+        const itemsCount = items.length;
+        const neededItemsCount = Math.min((activePage + 1) * perPage, totalCount);
+        const allNeededItemsLoaded = itemsCount >= neededItemsCount;
+
+        if (!allNeededItemsLoaded) {
+            return <LoadingSpinnerComponent extraClasses="c-content-table__loading-spinner" />;
+        }
+
+        if (!itemsCount) {
```
dew326: maybe we can do something like this?
```js
if (!itemsCount || !pages.hasOwnProperty(activePage)) {
    return null;
}

return pages[activePage].map(renderItem);
```

src/modules/common/popup/popup.component.js
```diff
@@ -15,161 +18,110 @@ const MODAL_SIZE_CLASS = {
...
-    }
-
-    setModalRef(component) {
-        this._refModal = component;
-    }
-
-    renderHeader() {
-        const closeBtnLabel = Translator.trans(/*@Desc("Close")*/ 'popup.close.label', {}, 'universal_discovery_widget');
-
+    }, []);
```
tischsoic: I think we should put `setIsVisible` as `useCallback` dependency.

PR #195, opened 2019-06-11
Add `style` prop on Icon component

tischsoic:
I want to pass custom inline styles on the Icon common component.

Neede for: https://github.com/ezsystems/ezplatform-calendar/pull/5


tischsoic:
We changed approach in styling action buttons in https://github.com/ezsystems/ezplatform-calendar/pull/5 and we no longer need changes from this PR.

PR #204, opened 2019-08-09
EZP-30526: Convert Search Tab as a standalone ReactJS module

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-30526

### DIFF
--- a/src/modules/ud/components/content-type-selector/content.type.selector.js
+++ b/src/modules/ud/components/content-type-selector/content.type.selector.js
@@ -1,0 +1,72 @@
+import React, { useState, useContext } from 'react';
+
+import { createCssClassNames } from '../../../common/helpers/css.class.names';
+import { SelectedContentTypesContext } from '../search/search';
+
+const ContentTypeSelector = () => {
+    const { contentTypes: contentTypesMap } = window.eZ.adminUiConfig;
+    const [selectedContentTypes, dispatchSelectedContentTypesAction] = useContext(SelectedContentTypesContext);
+    const [collapsedGroups, setCollapsedGroups] = useState(() => {
+        return Object.keys(contentTypesMap).reduce((collapsedGroups, contentTypeGroup, index) => {
+            return { ...collapsedGroups, [contentTypeGroup]: !!index };
+        }, {});
+    });
+    const toggleCollapsed = ({ nativeEvent }) => {
+        const contentTypeGroup = nativeEvent.target.dataset.id;
+
+        setCollapsedGroups((prevState) => ({ ...prevState, [contentTypeGroup]: !prevState[contentTypeGroup] }));
+    };
+    const handleContentTypeSelect = ({ nativeEvent }) => {
+        const contentTypeIdentifier = nativeEvent.target.dataset.contentTypeIdentifier;
+        const action = { contentTypeIdentifier };
+
+        action.type = selectedContentTypes.includes(contentTypeIdentifier) ? 'REMOVE_CONTENT_TYPE' : 'ADD_CONTENT_TYPE';
+
+        dispatchSelectedContentTypesAction(action);
+    };
+
+    return (
+        <div className="ez-content-type-selector c-content-type-selector">
+            {Object.entries(contentTypesMap).map(([contentTypeGroup, contentTypes]) => {
+                const groupSelectorClassName = createCssClassNames({
+                    'ez-content-type-selector__group': true,
+                    'ez-content-type-selector__group--collapsed': collapsedGroups[contentTypeGroup],
+                });
+
+                return (
+                    <div key={contentTypeGroup} className={groupSelectorClassName}>
+                        <span className="ez-content-type-selector__group-title" data-id={contentTypeGroup} onClick={toggleCollapsed}>
+                            {contentTypeGroup}
+                        </span>
+                        <ul className="ez-content-type-selector__list">
+                            {contentTypes.map((contentType) => {
+                                return (
+                                    <li key={contentType.identifier} className="ez-content-type-selector__item">
+                                        <div className="form-check form-check-inline">
+                                            <input
+                                                type="checkbox"
+                                                id={`ez-search-content-type-${contentType.identifier}`}
+                                                className="form-check-input"
+                                                value={contentType.identifier}
+                                                data-content-type-identifier={contentType.identifier}
+                                                onChange={handleContentTypeSelect}
+                                                checked={selectedContentTypes.includes(contentType.identifier)}
+                                            />
+                                            <label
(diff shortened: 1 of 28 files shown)


src/modules/ud/universal.discovery.module.js
```diff
@@ -29,8 +29,14 @@ const restInfo = {
     token: document.querySelector('meta[name="CSRF-Token"]').content,
     siteaccess: document.querySelector('meta[name="SiteAccess"]').content,
 };
+const contentTypesMap = Object.values(eZ.adminUiConfig.contentTypes).reduce((contentTypesMap, contentTypesGroup) => {
+    contentTypesGroup.forEach((contentType) => (contentTypesMap[contentType.href] = contentType));
```
tischsoic: Wrapping assignment `contentTypesMap[contentType.href] = contentType` into `()` is confusing for me. I would prefer standard `{}`.

src/modules/ud/services/universal.discovery.service.js
```diff
@@ -48,3 +48,73 @@ export const findLocationsByParentLocationId = (
...
+        method: 'POST',
+        headers: { ...HEADERS_CREATE_VIEW, 'X-Siteaccess': siteaccess, 'X-CSRF-Token': token },
+        body,
+        mode: 'same-origin',
+        credentials: 'same-origin',
+    });
+
+    fetch(request)
+        .then(handleRequestResponse)
+        .then((json) => {
```
tischsoic: Naming it `json` is a little bit odd, I would call it `response`.

src/modules/ud/components/toggle-selection-button/toggle.selection.button.js
```diff
@@ -0,0 +1,34 @@
...
+import PropTypes from 'prop-types';
+
+import Icon from '../../../common/icon/icon';
+
+import { createCssClassNames } from '../../../common/helpers/css.class.names';
+import { SelectedLocationsContext } from '../../universal.discovery.module';
+
+const ToggleSelectionButton = ({ location }) => {
+    const [selectedLocations, dispatchSelectedLocationsAction] = useContext(SelectedLocationsContext);
+    const isSelected = selectedLocations.some((selectedLocation) => selectedLocation.id === location.id);
```
tischsoic, line 11: Have you thought about using _map_ to speed up searching/removing?

src/modules/ud/components/search/search.js
```diff
@@ -0,0 +1,141 @@
...
+    const handleKeyPressed = ({ charCode }) => {
+        if (charCode === ENTER_CHAR_CODE) {
+            search(0);
+        }
+    };
+    const changePage = (pageIndex) => {
+        setOffset(pageIndex * limit);
+    };
+    const toggleFiltersCollapsed = () => {
+        setFiltersCollapsed((prevState) => {
```
tischsoic: Nitpick: we can do oneliner here: `setFiltersCollapsed((prevState) => !prevState);`

src/modules/ud/components/search/search.js
```diff
@@ -0,0 +1,141 @@
...
+                <ContentTable
+                    count={data.count}
+                    items={data.items}
+                    itemsPerPage={limit}
+                    activePageIndex={offset ? offset / limit : 0}
+                    title={title}
+                    onPageChange={changePage}
+                />
+            );
+        } else if (!!data.items) {
```
tischsoic, line 71: Hmm, why we need to check `!!data.items` condition?

src/modules/ud/components/filters/filters.js
```diff
@@ -0,0 +1,163 @@
...
+                    disabled={!isApplyButtonEnabled}>
+                    Apply
+                </button>
+            </div>
+        </div>
+    );
+};
+
+Filters.propTypes = {
+    collapsed: PropTypes.bool.isRequired,
```
tischsoic: following our convention: `isCollapsed`

src/modules/ud/components/filters/filters.js
```diff
@@ -0,0 +1,163 @@
...
+
+        const mergedConfig = {
+            onConfirm,
+            onCancel: closeUDW,
+            tabs: window.eZ.adminUiConfig.universalDiscoveryWidget.tabs,
+            title: 'Browsing content',
+            ...config,
+        };
+
+        ReactDOM.render(React.createElement(eZ.modules.UDW, mergedConfig), udwContainer);
```
tischsoic, line 73: Maybe better option would be to use `createPortal`?

src/modules/ud/components/filters/filters.js
```diff
@@ -0,0 +1,163 @@
...
+                <button type="button" className="ez-tag__remove-btn" onClick={clearSelectedSubree}>
+                    <Icon name="circle-close" extraClasses="ez-icon--medium ez-icon--dark" />
+                </button>
+            </div>
+        );
+    };
+
+    useEffect(() => {
+        if (filtersCleared) {
+            setFiltersCleared(false);
```
tischsoic, line 105: I'm not 100% what's the purpose of this construct, but AFAIU: we have separate state to trigger `makeSearch()` after state was updated. If that's the case I think it is anti-pattern. We should remove `filtersCleared` and add proper dependencies to `useEffect`.

src/modules/ud/components/content-type-selector/content.type.selector.js
```diff
@@ -0,0 +1,74 @@
...
+        }, {});
+    });
+    const toggleCollapsed = ({ nativeEvent }) => {
+        const contentTypeGroup = nativeEvent.target.dataset.id;
+
+        setCollapsedGroups((prevState) => {
+            return { ...prevState, [contentTypeGroup]: !prevState[contentTypeGroup] };
+        });
+    };
+    const handleContentType = ({ nativeEvent }) => {
```
tischsoic: This name does not tell what this function is handling. I would change it to something like: `handleContentTypeSelect` etc.

src/modules/ud/components/content-type-selector/content.type.selector.js
```diff
@@ -0,0 +1,74 @@
...
+                });
+
+                return (
+                    <div key={contentTypeGroup} className={groupSelectorClassName}>
+                        <span className="ez-content-type-selector__group-title" data-id={contentTypeGroup} onClick={toggleCollapsed}>
+                            {contentTypeGroup}
+                        </span>
+                        <ul className="ez-content-type-selector__list">
+                            {contentTypes.map((contentType) => {
+                                return (
```
tischsoic, line 43: Nitpick: here and in other places we can use `=> (/* val to return */)`. 😝

PR #208, opened 2019-08-28
EZP-30899: Change default CT sorting field from identifier to name in AdminUI

konradoboza:
JIRA: https://jira.ez.no/browse/EZP-30899

This PR adds the corresponding sorting in UDW.

related: https://github.com/ezsystems/ezplatform-admin-ui/pull/1072

### DIFF
--- a/src/modules/universal-discovery/components/create/choose.content.type.component.js
+++ b/src/modules/universal-discovery/components/create/choose.content.type.component.js
@@ -100,12 +100,20 @@ export default class ChooseContentTypeComponent extends Component {
     }
 
     renderGroup(groupName, index) {
-        const items = this.props.contentTypes[groupName];
         const groupAttrs = {};
+
+        let items = this.props.contentTypes[groupName];
 
         if (this.state.filterQuery && items.every((item) => !item.name.toLowerCase().includes(this.state.filterQuery))) {
             groupAttrs.hidden = true;
         }
+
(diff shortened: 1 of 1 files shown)


dew326:
Why the sorting is a problem in the content on the fly but it is not a problem in the create widget in content structure or in the content type list in the search filter?

In my opinion, if the user gets used to that article is on the 10th position in the create widget in the content structure he will be looking article at the same position in the content on the fly.

If we want to change the order of content types we should change this across the whole system. 
This should be done on the backend when the content types are set in the adminUiConfig or even deeper in the kernel.

konradoboza:
My bad, I haven't mentioned the second PR https://github.com/ezsystems/ezplatform-admin-ui/pull/1072. As far as I can see it changes the order of the Content Types accordingly in all places across the system. Are you able to point me to the place which isn't covered by this change? I fully agree that it needs to be consistent across the system, that was my initial intention.

dew326:
Sorry, not really I don't know where this is used and where is not.

But still, when it comes to frontend this should be sorted in the `adminUiConfig.contentTypes`. We should sort on the input, not in the usage.

konradoboza:
Sorting in UDW and other places handled in: https://github.com/ezsystems/ezplatform-admin-ui/pull/1072/commits/8abef1c5208adc887f597e5f48e31e408640bb95, closing this one as obsolete.

PR #222, opened 2019-10-21
EZP-30159: Content can't be created in the preconfigured location using UDW.

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-30159

Now the content on the fly (create tab) will render additional branch with the starting location except the root location with id:1 (because in the whole application there is no way to create content under the root location).

**Screen:**
cotf-starting-location
On the screen the starting location is set to `Images` folder


katarzynazawada (approved):

lserwatka:
Could you merge it up?

Issue #230, opened 2019-11-27
UDW in ezobjectrelationlist: Search Component does not take ctofAllowedContentType config


maczuk:
I'm searching for content in ezobjectrelationlist field type with content type limitation set. UDW does not pass this configuration from `UniversalDiscoveryModule` to `SearchComponent`. Because of that search results have contents that can not be used (wrong content type).

Tech details:
**1) field of ezobjectrelationlist field type with configured content type limitation**
![eZ Platform 2019-11-27 12-03-39](https://user-images.githubusercontent.com/5550410/69718254-2f5c1680-110e-11ea-90ce-e5d89e142ed9.png)

**2)  query only with search text but no content type limitation -> search results with tons of contents of invalid content type**
![eZ Platform 2019-11-27 12-08-10](https://user-images.githubusercontent.com/5550410/69718879-a645df00-110f-11ea-8089-47951a5dcb46.png)

maczuk:
I have created PR for this [PR#231](https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/231)

What do you think?

dew326:
Hi, we are using Jira. Ticket for this issue is already created: https://jira.ez.no/browse/EZP-31194

PR #231, opened 2019-11-27
EZP-31194: Searching in UDW does not take allowed CT's into account when ezobjectrelation FieldType is used

maczuk:
Fix for issue [#230 UDW in ezobjectrelationlist: Search Component does not take ctofAllowedContentType config](https://github.com/ezsystems/ezplatform-admin-ui-modules/issues/230)

Jira ticket: https://jira.ez.no/browse/EZP-31194
Requires: https://github.com/ezsystems/ezplatform-admin-ui/pull/1145

For QA:
Please check also this flow:

1. Configure `allowed_content_types` to ['article'] for i.e browse button
2. Configure `content_on_the_fly.allowed_content_types` to ['blog_post'] 

Expected:
In the content on the fly (create tab) only 'blog_post' is allowed, in other tabs only 'article' can be selected


src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -901,6 +902,7 @@ export default class UniversalDiscoveryModule extends Component {
             bookmarksPerPage,
             restInfo,
             selectedContent,
+            allowedContentTypes: cotfAllowedContentTypes,
```
dew326: Here, the `allowedContentTypes` should be passed not `cotfAllowedContentTypes`.

src/modules/universal-discovery/components/search/search.component.js
```diff
@@ -37,12 +37,19 @@ export default class SearchComponent extends Component {
         }
 
         const searchText = this._refSearchInput.value;
+        const query = {
+            FullTextCriterion: searchText
```
dew326:
```suggestion
            FullTextCriterion: searchText,
```

dew326 (changes requested):
I had this issue in my queue, but you were faster, thanks for the contribution :)

Here is the Jira ticket for the issue: https://jira.ez.no/browse/EZP-31194
Can you update the PR title and commit message?

src/modules/universal-discovery/universal.discovery.module.js
maczuk: in HTML attribute `data-udw-config` allowed content types are in `ctofAllowedContentTypes` (look in this [prtcsr](https://user-images.githubusercontent.com/5550410/69718254-2f5c1680-110e-11ea-90ce-e5d89e142ed9.png)) and in line you commented I map value of this key to `allowedContentTypes` key and next pass to `SearchPanelComponent` with `allowedContentTypes` key (not `ctofAllowedContentTypes`) (look in `search.component.js` line 45: `query.ContentTypeIdentifierCriterion = this.props.allowedContentTypes;`)

dew326: This key is only for content on the fly, all other tabs should use the `allowedContentTypes`. I know this might be confusing but it was done for the BC.

maczuk:
@dew326 - Now it's good, I think ;)

dew326:
You have one commit too much https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/231/commits/b3629614e2bdefc9922134f82f65246be7100c63
And still `cotfAllowedContentTypes` should be changed to `allowedContentTypes`

maczuk:
Yeah, sorry! I removed this commit.

But I don't understand this `[cotf]allowedContentType` problem :) 

Look at this [prtscr](https://user-images.githubusercontent.com/5550410/69718254-2f5c1680-110e-11ea-90ce-e5d89e142ed9.png) - in config (json string in `data-udw-config` attribute) only `cotfAllowedContentTypes` is set, `allowedContentTypes` key has empty array. I have also log `props` of `UniversalDiscoveryModule` and recieved the same result - `cotfAllowedContentTypes` array is filled, but `allowedContentTypes` has empty array... look:

good

so if I would use `allowedContentTypes` as you say, empty array would be passed to `SearchComponent` and it would not work

also, I found in docs [Universal Discovery Widget (UDW) - Optional props](https://doc.ezplatform.com/en/2.5/guide/extending/extending_modules/#optional-props) info about `cotfAllowedContentTypes` and there is no info regarding `allowedContentTypes`...

dew326:
I agree we are missing information about `allowedContentTypes`.

From the doc: cotfAllowedContentTypes {Array} - Content Types that are available in Content on the Fly. We are fixing this issue in documentation in this PR https://github.com/ezsystems/developer-documentation/pull/877

In your current configuration you will be able to create only `keyword` content but select all contents in browse tab. If you configure the `allowedContentTypes` array (https://github.com/ezsystems/ezplatform-admin-ui/blob/1.5/src/bundle/Resources/config/universal_discovery_widget.yml#L13), you will not be able to select other content than `keyword` - this is proper behavior.

src/modules/universal-discovery/services/universal.discovery.service.js
```diff
@@ -193,7 +193,7 @@ export const findLocationsByParentLocationId = (
  *
  * @function findContentBySearchQuery
  * @param {Object} restInfo REST config hash containing: token and siteaccess properties
- * @param {String} query
+ * @param {Object} query Object containing query criterions
```
dew326, line 196: Also, the `identifier` in `ViewInput` should be changed:
`udw-locations-by-search-query-${query}` => `udw-locations-by-search-query-${query.FullTextCriterion}`

maczuk:
Yes, this is proper behaviour and it works as you described. If I'm not using cotfAllowedContentTypes I can't narrow my search results and I have to scroll through tons of content which I can't select. 
Whats the motivation to look up for a content I'm not able to use?

Additionaly, maybe `ezobjectrelationlist` Field Type has a bug, and does not fill `allowedContentTypes` key:
eZ Platform 2019-11-27 15-37-39

maczuk:
@dew326 I made the changes we talked about yesterday at Slack.

I also created PR for ezsystems/ezplatform-admin-ui bugfix: [ezsystems/ezplatform-admin-ui#1145](https://github.com/ezsystems/ezplatform-admin-ui/pull/1145)

src/modules/universal-discovery/universal.discovery.module.js
```diff
@@ -820,14 +820,14 @@ export default class UniversalDiscoveryModule extends Component {
...
                 contentTypes,
+                allowedContentTypes,
                 onLanguageSelected: this.onLanguageSelected,
                 onContentTypeSelected: this.onContentTypeSelected,
                 contentTypesMap: this.state.contentTypesMap,
                 forcedLanguage: cotfForcedLanguage,
                 preselectedLanguage: cotfPreselectedLanguage,
                 allowedLanguages: cotfAllowedLanguages,
                 preselectedContentType: cotfPreselectedContentType,
-                allowedContentTypes: cotfAllowedContentTypes,
```
dew326: This is the configuration for content on the fly (create tab) and here should be `cotfAllowedContentTypes` passed. It was ok.

src/modules/universal-discovery/universal.discovery.module.js
dew326: This one should be changed to `allowedContentTypes`

dew326 (changes requested):

dew326 (approved):

micszo (changes requested):
When a CT restriction is set on ezobjectrelation field and content_on_the_fly.allowed_content_types is set to some CT the result in UDW on Create tab should be a intersection / common part not a sum (ref. https://jira.ez.no/browse/EZP-31089)

maczuk:
> When a CT restriction is set on ezobjectrelation field and content_on_the_fly.allowed_content_types is set to some CT the result in UDW on Create tab should be a intersection / common part not a sum (ref. https://jira.ez.no/browse/EZP-31089)

I made this fix here [ezsystems/ezplatform-admin-ui#1145](https://github.com/ezsystems/ezplatform-admin-ui/pull/1145)

micszo (approved):
QA Approved on eZ Platform EE v2.5.7 with diffs.
For the intersection issue patches from EZP-31089 have to be applied also (or v2.5.8 will have to be used).

lserwatka:
@dew326 could you merge it up.

PR #234, opened 2019-12-02
EZP-31089: Wrong allowed content types list on object relation(s) fields

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-31089


micszo (commented):
Works fine, tested together with https://github.com/ezsystems/ezplatform-admin-ui/pull/1155.

micszo (approved):

PR #250, opened 2020-01-17
EZEE-2956: Auto confirm is not enabled in UDW for Embed block

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZEE-2956
**Requires:** https://github.com/ezsystems/ezplatform-admin-ui/pull/1200

For QA:
With auto-confirm and multiple settings, UDW should return both selected items and just create one. With single should return only the create item.

Without auto-confirm and multiple after creation item should be added to the list (if the limit is reached the create button should be disabled). WIth a single creation should replace the selected item.


OstafinL (approved):

GrabowskiM (approved):

micszo (approved):
QA Approved on eZ Platform EE master.
Follow-up tickets in related PR.

PR #291, opened 2021-06-07
IBX-445: selected first enabled language for create content

mateuszdebinski:
jira ticket: https://issues.ibexa.co/browse/IBX-445

### DIFF
--- a/src/modules/universal-discovery/components/create/choose.language.component.js
+++ b/src/modules/universal-discovery/components/create/choose.language.component.js
@@ -9,7 +9,7 @@ export default class ChooseLanguageComponent extends Component {
         this.renderOption = this.renderOption.bind(this);
 
         let selectedLanguageCode = null;
-        let selectedLanguage;
+        let selectedLanguage = {};
 
         if (props.preselectedLanguage) {
             selectedLanguageCode = props.preselectedLanguage;
@@ -35,15 +35,16 @@ export default class ChooseLanguageComponent extends Component {
                 return isAllowedLanguage && language.enabled;
             });
 
-            selectedLanguage = language || {};
+            if (language !== undefined) {
+                selectedLanguage = language;
+            }
         } else {
-            selectedLanguage = props.languages.mappings[selectedLanguageCode]
+            selectedLanguage = props.languages.mappings[selectedLanguageCode];
         }
 
         this.state = {
-            selectedLanguage: selectedLanguage,
+            selectedLanguage,
         };
-
     }
 
     componentDidMount() {
--- a/src/modules/universal-discovery/components/create/choose.language.component.js
+++ b/src/modules/universal-discovery/components/create/choose.language.component.js
@@ -70,7 +70,7 @@ export default class ChooseLanguageComponent extends Component {
             return null;
         }
 
-        if (this.state.selectedLanguage !== undefined && this.state.selectedLanguage.languageCode === languageCode) {
+        if (this.state.selectedLanguage.languageCode === languageCode) {
             attrs.selected = true;
         }
 
--- a/src/modules/universal-discovery/components/create/choose.language.component.js
+++ b/src/modules/universal-discovery/components/create/choose.language.component.js
@@ -9,6 +9,7 @@ export default class ChooseLanguageComponent extends Component {
         this.renderOption = this.renderOption.bind(this);
 
         let selectedLanguageCode = null;
+        let selectedLanguage;
 
         if (props.preselectedLanguage) {
             selectedLanguageCode = props.preselectedLanguage;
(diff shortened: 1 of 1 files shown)


src/modules/universal-discovery/components/create/choose.language.component.js
```diff
@@ -26,6 +26,17 @@ export default class ChooseLanguageComponent extends Component {
...
 
+        if (selectedLanguageCode === null || !props.languages.mappings[selectedLanguageCode].enabled) {
+            const languages = Object.values(props.languages.mappings);
+            const filteredLanguages = languages.filter((language) => {
+                const isAllowedLanguage = props.allowedLanguages.length ? props.allowedLanguages.includes(language.languageCode) : true;
+
+                return isAllowedLanguage && language.enabled;
+            });
+
+            selectedLanguageCode = filteredLanguages.length ? filteredLanguages[0].languageCode : '';
```
dew326: I think if `selectedLanguageCode` will be `''` the UDW will crash.

dew326 (changes requested):

src/modules/universal-discovery/components/create/choose.language.component.js
mateuszdebinski: it is not possible, because the only one where language is verified via the state for content creation is in the `renderOptions` function, but it will not be performed as well because no option will be available or enabled and form can be submitted

dew326: After this component is mouted it updates the selectedLanguage in parent component. The parent component checks permissions and may do other stuff. It would be safer to never set `selectedLanguage` to null, better would be `{}`.

src/modules/universal-discovery/components/create/choose.language.component.js
```diff
@@ -26,6 +26,17 @@ export default class ChooseLanguageComponent extends Component {
             selectedLanguageCode = props.allowedLanguages[0];
         }
 
+        if (selectedLanguageCode === null || !props.languages.mappings[selectedLanguageCode].enabled) {
+            const languages = Object.values(props.languages.mappings);
+            const filteredLanguages = languages.filter((language) => {
```
dew326: You can use `find` to not have to go through the whole array, just stop on the first hit.

dew326 (changes requested):

OstafinL (approved):

src/modules/universal-discovery/components/create/choose.language.component.js
```diff
@@ -26,9 +27,23 @@ export default class ChooseLanguageComponent extends Component {
...
+            const languages = Object.values(props.languages.mappings);
+            const language = languages.find((language) => {
+                const isAllowedLanguage = props.allowedLanguages.length ? props.allowedLanguages.includes(language.languageCode) : true;
+
+                return isAllowedLanguage && language.enabled;
+            });
+
+            selectedLanguage = language || {};
+        } else {
+            selectedLanguage = props.languages.mappings[selectedLanguageCode]
```
dew326:
```suggestion
            selectedLanguage = props.languages.mappings[selectedLanguageCode];
```

src/modules/universal-discovery/components/create/choose.language.component.js
```diff
@@ -26,9 +27,23 @@ export default class ChooseLanguageComponent extends Component {
...
+        }
+
         this.state = {
-            selectedLanguage: props.languages.mappings[selectedLanguageCode],
+            selectedLanguage: selectedLanguage,
```
dew326:
```suggestion
            selectedLanguage,
```

src/modules/universal-discovery/components/create/choose.language.component.js
```diff
@@ -26,9 +27,23 @@ export default class ChooseLanguageComponent extends Component {
...
         };
+
```
dew326:
```suggestion
```

src/modules/universal-discovery/components/create/choose.language.component.js
```diff
@@ -8,7 +8,8 @@ export default class ChooseLanguageComponent extends Component {
         this.updateSelection = this.updateSelection.bind(this);
         this.renderOption = this.renderOption.bind(this);
 
-        let selectedLanguageCode = props.allowedLanguages.length ? props.allowedLanguages[0] : props.languages.priority[0];
+        let selectedLanguageCode = null;
+        let selectedLanguage;
```
dew326:
```suggestion
        let selectedLanguage = {};
```

src/modules/universal-discovery/components/create/choose.language.component.js
```diff
@@ -26,9 +27,23 @@ export default class ChooseLanguageComponent extends Component {
...
 
+        if (selectedLanguageCode === null || !props.languages.mappings[selectedLanguageCode].enabled) {
+            const languages = Object.values(props.languages.mappings);
+            const language = languages.find((language) => {
+                const isAllowedLanguage = props.allowedLanguages.length ? props.allowedLanguages.includes(language.languageCode) : true;
+
+                return isAllowedLanguage && language.enabled;
+            });
+
+            selectedLanguage = language || {};
```
dew326: I would set it only if the `language` exists.
Thanks to the changes `selectedLanguage` is never `null`

dew326 (changes requested):

dew326 (approved):

bogusez (approved):

adamwojs:
Could you please merge up changes @mateuszdebinski?

